PR #5, opened 2023-04-17
Translator for lightkurve.LightCurve objects to glue Data objs

bmorris3:
This PR has an implementation for a `glue-astronomy`-style translator from `lightkurve.LightCurve` objects to `glue.core.Data` objects. The contents of the `lcviz/utils.py` file will ultimately go into a PR to `glue-astronomy`, but I'm putting them here for now, so we don't have to wait on upstream releases to work here. 

My assigned ticket was to actually parse Kepler FITS files, which will be trivial after this first-step PR is merged, since those features are implemented in lightkurve. In the soon-to-come follow-up PR, I'll implement something like a `lightkurve_parser` which uses lightkurve's [`read` function](https://docs.lightkurve.org/reference/io.html) to construct a `LightCurve` from a Kepler or TESS data product, and rely on the translator in this PR to send the contents into glue `Data`.

[🐱](https://jira.stsci.edu/browse/JDAT-3143)

### DIFF
--- a/lcviz/__init__.py
+++ b/lcviz/__init__.py
@@ -5,6 +5,8 @@ try:
 except ImportError:
     __version__ = ''
 
+from . import utils  # noqa
+
 # Expose subpackage API at package level.
 from .plugins import *  # noqa
 from .parsers import *  # noqa
--- a/lcviz/conftest.py
+++ b/lcviz/conftest.py
@@ -1,0 +1,60 @@
+import numpy as np
+import pytest
+from astropy import units as u
+from astropy.utils.masked import Masked
+from lightkurve import LightCurve
+
+from lcviz import __version__
+
+
+@pytest.fixture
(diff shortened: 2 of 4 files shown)


bmorris3:
I'm not sure it makes to test subsets sense until we decide _how_ to define subsets in time or flux... thoughts?

kecnry:
that's a good point.  I think we'll eventually want xrange, yrange, and box/mask subsets... but maybe just a simple time xrange subset (if it can even be triggered from the API at this point) would be enough to get the tests into that codeblock?  Otherwise, let's create a follow-up issue/ticket to write tests that go through the translator for ALL supported subsets that we eventually enable?

bmorris3:
@kecnry I've started implementing a test with a subset in time, but there are no subset tools for the helper yet – you can select an x-range manually or with `XRangeROI` – neither selects a subset.

PR #7, opened 2023-04-18
jdaviz spectrum viewer inheritance

kecnry:
This PR "fixes" (in a really hacky way) some missing methods that we need from jdaviz's spectrum viewer that are not available in the `JdavizViewerMixin` mixin that `TimeProfileView` inherits.  We don't want to inherit directly from `SpecvizProfileView`, because that comes with a lot of spectrum-specific methods that we don't want here.  Ideally, we should refactor this on the jdaviz-end to pull these shared methods into a separate mixin which can then be used from lcviz.

### DIFF
--- a/lcviz/viewers.py
+++ b/lcviz/viewers.py
@@ -6,6 +6,8 @@ from glue_jupyter.bqplot.profile import BqplotProfileView
 
 from jdaviz.core.registries import viewer_registry
 from jdaviz.configs.default.plugins.viewers import JdavizViewerMixin
+from jdaviz.configs.specviz.plugins.viewers import SpecvizProfileView
+from jdaviz.core.freezable_state import FreezableProfileViewerState
 
 
 @viewer_registry("lcviz-time-viewer", label="Profile 1D (LCviz)")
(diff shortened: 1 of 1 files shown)


kecnry:
I think my vote would be to move forward with this hack to avoid tracebacks in the plot options plugin and zoom tools, open up a ticket in jdaviz to pull generic methods out of the `SpecvizProfileView` into a new mixin (and then inherit that mixin here and remove the hack), as well as a ticket to either generalize uncertainty plotting in jdaviz or write our own here.  Anyone have any opinions on that plan?

lcviz/viewers.py
```diff
@@ -18,10 +20,18 @@ class TimeProfileView(JdavizViewerMixin, BqplotProfileView):
...
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.initialize_toolbar()
+        self._subscribe_to_layers_update()
+        # hack to inherit a small subset of methods from SpecvizProfileView
+        # TODO: refactor jdaviz so these can be included in some mixin
+        self._show_uncertainty_changed = lambda value: SpecvizProfileView._show_uncertainty_changed(self, value)  # noqa
+        self._plot_uncertainties = lambda: SpecvizProfileView._plot_uncertainties(self)
+        # TODO: _plot_uncertainties in specviz is hardcoded to look at spectral_axis and so crashes
```
bmorris3, line 33: Should we prevent users from plotting uncertainties for now to avoid the crashes?

bmorris3 (approved):

lcviz/viewers.py
kecnry: I'm hoping we can fix this properly soon.  If we don't, then let's definitely have that conversation of the best way to disable it (there are multiple places in the API and plot options UI that are affected).

PR #10, opened 2023-04-18
First light curve parser

bmorris3:
This PR implements a first crack at the light curve parser.

If you have a local Kepler FITS data product (or can download one from MAST), you can load it into LCviz with:

```python
from lcviz import LCviz

light_curve = search_lightcurve(
    "HAT-P-11", mission="Kepler", cadence="long", quarter=10
).download()

lcviz = LCviz()

# # Optionally normalize the light curve:
# light_curve = light_curve.normalize()

lcviz.load_data(light_curve)
lcviz.show()
```
Screen Shot 2023-04-18 at 17 02 09

I've made the x label show the time since the start of the reference data, and it defaults to units of seconds (the WCS default, I think?), and the y label shows units of "Flux" for count rates, and "Relative flux" for dimensionless light curves. 

I've written some tests that I had anticipated would fail because of https://github.com/lightkurve/lightkurve/pull/1299, but it looks like that error may have been a mirage from a mismatch of the astropy and lightkurve versions (still investigating).


kecnry (approved):
Not sure how I feel about the time since reference thing... we'll likely eventually want options for that in a plugin somewhere... but works for now!  Thanks!

bmorris3:
> Not sure how I feel about the time since reference thing... 

I agree it's not perfect. I usually prefer to plot my transiting exoplanet LCs in units of `BJD - <some number>`, [like this](https://www.aanda.org/articles/aa/full_html/2021/09/aa40892-21/F4.html). In general of course, LCs can span only a few seconds, or many years, so having unit options other than seconds would be great.

> we'll likely eventually want options for that in a plugin somewhere... 

Agreed. This might require a bit of work. I'm using the `ndcube` [TimeTableCoordinate](https://docs.sunpy.org/projects/ndcube/en/stable/api/ndcube.extra_coords.table_coord.TimeTableCoordinate.html#ndcube.extra_coords.table_coord.TimeTableCoordinate) to convert `Time` objects into gwcs representations of the time axis. `TimeTableCoordinate` defaults to units of seconds, I think because the WCS TIME spec expects seconds. More to look into there.

bmorris3:
With approval from @kecnry, I'll merge now and make a ticket to follow-up when https://github.com/lightkurve/lightkurve/pull/1299 is addressed. Thanks!

PR #11, opened 2023-04-19
move build configuration from `setup.cfg` to `pyproject.toml`

zacharyburnett:
`setuptools` now supports the `[project]` table, which is defined by [PEP621](https://peps.python.org/pep-0621).

Additionally, `setuptools` now supports its own entry in `pyproject.toml` called `[tool.setuptools]` (https://github.com/pypa/setuptools/issues/1688, https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration); however, it comes with the following disclaimer:
> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the `[tool.setuptools]` table), is still in **beta** stage and might change in future releases.

Reading `toml` is supported natively in Python 3.11 with `tomllib`

Given this, we can attempt to consolidate the build configuration into a single `pyproject.toml` file that can possibly be read by other build systems in the future.


duytnguyendtn (approved):
Installs and imports nicely on my side. Thanks for the modernization effort!

pyproject.toml
```diff
@@ -1,5 +1,131 @@
...
+    "except ImportError",
+    "raise AssertionError",
+    "raise NotImplementedError",
+    "def main\\(.*\\):",
+    "pragma: py{ignore_python_version}",
+    "def _ipython_key_completions_",
+]
+
+[tool.build_sphinx]
+edit-on-github = true
```
rosteen, line 130:
```suggestion
edit_on_github = true
```
This made the `readthedocs` build fail in `jdaviz`, I assume it would be the same here. Surprisingly nothing seemed to fail due the the line below being dashes instead of underscores, but maybe we should keep it in mind in case some link doesn't work.

pyproject.toml
```diff
@@ -1,5 +1,131 @@
...
+github-project = "spacetelescope/lcviz"
```
rosteen, line 131:
```suggestion
github_project = "spacetelescope/lcviz"
```
Actually I lied, turns out it is an underscore in `jdaviz`.

rosteen (approved):

PR #12, opened 2023-04-27
Adding support for subsets in time

bmorris3:
This PR gets time subsets working for LCviz. [🐱](https://jira.stsci.edu/browse/JDAT-3297)

https://user-images.githubusercontent.com/3497584/234955643-55aeb31c-5a7e-4283-94a1-0fbc1bbd7ca9.mov

Covered in this PR:
* Added a Kepler/TESS quality flag component to the translated `LightCurve`s (will later useful for masking)
* adapted the `LightCurve` translator in lcviz to enable subset selections from light curves
    * Made a new glue `Coordinates` subclass to use for the time axis (analog of glue-astro’s `SpectralCoordinates`)
* Added translators for the `KeplerLightCurve` and `TessLightCurve` subclasses of `LightCurve` (so the type of the input `*LightCurve` object is preserved on round-trip)

This PR depends on upstream updates to jdaviz as of https://github.com/spacetelescope/jdaviz/pull/2168, which adds support for temporal subset selection.

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -37,7 +37,10 @@ class LCviz(ConfigHelper):
                      'dense_toolbar': False,
                      'context': {'notebook': {'max_height': '600px'}}},
         'toolbar': ['g-data-tools', 'g-subset-tools'],
-        'tray': ['g-metadata-viewer', 'g-plot-options', 'g-subset-plugin', 'HelloWorldPlugin', 'g-export-plot'],
+        'tray': [
+            'g-metadata-viewer', 'g-plot-options', 'g-subset-plugin',
+            'HelloWorldPlugin', 'g-export-plot'
+        ],
         'viewer_area': [{'container': 'col',
                          'children': [{'container': 'row',
                                        'viewers': [{'name': 'time-viewer',
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -1,7 +1,30 @@
-from jdaviz.core.helpers import ConfigHelper
+import astropy.units as u
+
 from lightkurve import LightCurve
 
+from jdaviz.core.helpers import ConfigHelper
+
 __all__ = ['LCviz']
+
+
+def _get_range_subset_bounds(self, subset_state, *args, **kwargs):
+    # Instead of overriding the jdaviz version of this method on jdaviz.Application,
+    # we could put in jdaviz by (1) checking if helper has a
+    # _default_time_viewer_reference_name, (2) using the LCviz version if so, and (3)
+    # using the jdaviz version otherwise.
+    viewer = self.get_viewer(self._jdaviz_helper._default_time_viewer_reference_name)
+    reference_time = viewer.state.reference_data.meta['reference_time']
+    if viewer:
+        units = u.Unit(viewer.state.x_display_unit)
+    else:
+        raise ValueError("Unable to find time axis units")
+
+    region = reference_time + u.Quantity([subset_state.lo * units, subset_state.hi * units])
+
+    return [{"name": subset_state.__class__.__name__,
+             "glue_state": subset_state.__class__.__name__,
+             "region": region,
+             "subset_state": subset_state}]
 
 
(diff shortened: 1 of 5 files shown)


bmorris3:
The failing test is coming from the `jdaviz==3.4.0` pin. The implementation of temporal subsets here depends on the `get_subsets` implementation in jdaviz on main. The failure at `get_subsets_from_viewer` only occurs before @javerbukh's PR https://github.com/spacetelescope/jdaviz/pull/2087 was merged into main, which is milestoned for release v3.5. If/when we update the jdaviz pin to `jdaviz==3.5`, that error should disappear.

lcviz/helper.py
```diff
@@ -24,6 +51,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._default_time_viewer_reference_name = 'time-viewer'
 
+        # override jdaviz behavior to support temporal subsets
+        self.app._get_range_subset_bounds = (
+            lambda *args, **kwargs: _get_range_subset_bounds(self.app, *args, **kwargs)
+        )
```
kecnry, line 57: equally as hacky as #7 🤩 

Do you think there is anything we _should_ do upstream in the future that could get rid of this?

lcviz/viewers.py
```diff
@@ -81,3 +120,55 @@ def set_plot_axes(self):
...
+        # and all subsets affected will be spectral
+        for layer in self.state.layers:
+            if "Subset" in layer.layer.label and layer.layer.data.label == data.label:
+                layer.linewidth = 3
+
+        return result
+
+    def _show_uncertainty_changed(*args, **kwargs):
+        # method required by jdaviz
+        pass
```
kecnry, line 164: Let's create a follow-up issue to fix/remove this

lcviz/viewers.py
```diff
@@ -36,19 +49,45 @@ def __init__(self, *args, **kwargs):
...
+
+                    if _class is not None:
+                        handler, _ = data_translator.get_handler_for(_class)
+                        try:
+                            layer_data = handler.to_object(layer_data)
+                        except IncompatibleAttribute:
+                            continue
+                    data.append(layer_data)
+
+        return data
```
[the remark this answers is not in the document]
bmorris3: Agreed. Maybe we should add a comment saying something like "TODO: generalize upstream," and make a jdaviz techdebt ticket?

lcviz/helper.py
bmorris3: We could do what I suggest in [the comment just above this line](https://github.com/spacetelescope/lcviz/pull/12/files#diff-b6950a44b3dc9c628f86cb2931fb48fe2f81b46acab584c77f7228d13a2411c3R11-R14) 🙃

kecnry: wow, completely missed that, sorry!  That feels very lcviz-specific logic in jdaviz though... I actually think I prefer it the way you have it now, and if other custom configs motivate generalizing furthern, we can reconsider.

lcviz/viewers.py
bmorris3: Ticket added [👍🏻](https://jira.stsci.edu/browse/JDAT-3341)

kecnry (approved):
> Upstream fixes to jdaviz have been merged (https://github.com/spacetelescope/jdaviz/pull/2168), will be released in 3.4.1.

If the tests pass for you locally, then I'd be ok with bumping the pin here to 3.4.1 (so the tests fail saying there is no 3.4.1 until there actually is, rather than other failures caused by 3.4.0 and then later having to remember to bump the pin), and merging.

duytnguyendtn (approved):
Good to see subsets working again. Thanks Brett!

bmorris3:
Merging despite test failures. We'll need to update the jdaviz pin to 3.5 after that gets released to pick up the changes in https://github.com/spacetelescope/jdaviz/pull/2168 which support this PR.

PR #15, opened 2023-05-11
support for markers instead of lines via ScatterViewer

kecnry:
This PR swaps out the ProfileViewer with ScatterViewer so that we can view the light curve as markers instead of lines and eventually have different arrays along the x-axis (for phase-viewer support).

This also overrides the `user_api` for the plot options plugin, but requires https://github.com/spacetelescope/jdaviz/pull/2193 in order to have access to the necessary glue-state objects and UI for those entries in the plugin itself.  This can be tested against that PR, but will eventually need a bump in the version of jdaviz once that PR is finalized/released.  If for some reason the decision is made to not merge that into jdaviz, the plot options logic can be added to lcviz as a follow-up PR.

With the exception of plot options, everything else should work with jdaviz main.

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -13,9 +13,12 @@ def _get_range_subset_bounds(self, subset_state, *args, **kwargs):
     # _default_time_viewer_reference_name, (2) using the LCviz version if so, and (3)
     # using the jdaviz version otherwise.
     viewer = self.get_viewer(self._jdaviz_helper._default_time_viewer_reference_name)
-    reference_time = viewer.state.reference_data.meta['reference_time']
+    light_curve = viewer.data()[0]
+    reference_time = light_curve.meta['reference_time']
     if viewer:
-        units = u.Unit(viewer.state.x_display_unit)
+        # TODO: use display units once implemented in Glue for ScatterViewer
+        # units = u.Unit(viewer.state.x_display_unit)
+        units = u.Unit(viewer.time_unit)
     else:
         raise ValueError("Unable to find time axis units")
 
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -43,13 +43,13 @@ class LCviz(ConfigHelper):
         ],
         'viewer_area': [{'container': 'col',
                          'children': [{'container': 'row',
-                                       'viewers': [{'name': 'time-viewer',
+                                       'viewers': [{'name': 'flux-vs-time',
                                                     'plot': 'lcviz-time-viewer',
-                                                    'reference': 'time-viewer'}]}]}]}
+                                                    'reference': 'flux-vs-time'}]}]}]}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._default_time_viewer_reference_name = 'time-viewer'
+        self._default_time_viewer_reference_name = 'flux-vs-time'
 
         # override jdaviz behavior to support temporal subsets
         self.app._get_range_subset_bounds = (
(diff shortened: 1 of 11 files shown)


lcviz/parsers.py
```diff
@@ -43,6 +44,12 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
         new_data_label += f'[{flux_origin}]'
 
     app.add_data(light_curve, new_data_label)
+    dc = app.data_collection
+    if len(dc) > 1:
+        # then we need to link this light curve back to the first
+        dc0_comps = {str(comp): comp for comp in dc[0].components}
+        new_links = [LinkSame(dc0_comps.get(str(new_comp)), new_comp) for new_comp in dc[-1].components]  # noqa
```
kecnry, line 51: does anyone know of a cleaner way to do this reliably?  I do something similar but with `.index` later which might be a little cheaper, but both feel clunky.

lcviz/helper.py
```diff
@@ -38,18 +41,18 @@ class LCviz(ConfigHelper):
...
+            'g-metadata-viewer', 'lcviz-plot-options', 'g-subset-plugin',
             'HelloWorldPlugin', 'g-export-plot'
         ],
         'viewer_area': [{'container': 'col',
                          'children': [{'container': 'row',
-                                       'viewers': [{'name': 'time-viewer',
+                                       'viewers': [{'name': 'flux-vs-time',
                                                     'plot': 'lcviz-time-viewer',
-                                                    'reference': 'time-viewer'}]}]}]}
+                                                    'reference': 'flux-vs-time'}]}]}]}
```
bmorris3, line 51: Is there a y-quantity we plan to display other than "flux"? If not, I'd advocate for dropping `flux-vs-`.

lcviz/parsers.py
bmorris3: I suspect the efficiency difference of `get` vs `index` is small compared to the  linking expense, so this is probably fine.

Should we add a toggle like `imviz.load_data(... do_link=True)` to optionally turn off this block? There are use cases for separate LCs/viewers that shouldn't be linked.

lcviz/helper.py
kecnry: Eventually I think we'll have model parameter vs cycle/time (for time-dependent fitting of template models fitted in phase-space) and want the user to be able to create their own viewers (for uncertainty vs flux, or whatever)..... but I suppose that doesn't necessarily require the prefix here.  Until we do an actual release with example notebooks, this should be very easy to change and I think we'll get a better feel for what makes the most sense as we flesh out a few more features 🤞 so unless you feel strongly, let's leave it for now and revisit then.

rosteen:
I tried this with https://github.com/spacetelescope/jdaviz/pull/2193 and although the option to turn on the line mark is there in plot options, I don't actually see a line appear when I activate it. Is there something else I need for this, or is that expected behavior for now?

kecnry:
> the option to turn on the line mark is there in plot options, I don't actually see a line appear when I activate it. Is there something else I need for this, or is that expected behavior for now?

Right, glue-jupyter doesn't actually support plotting the lines yet.  Maybe we shouldn't even expose the switch until it does? 
 (that change should happen in https://github.com/spacetelescope/jdaviz/pull/2193 and would automatically be seen by lcviz).

rosteen (approved):
Given that all the marker options worked, and hiding the line toggle would be upstream anyway, I'm approving.

kecnry:
Holding off on merge until the upstream PR gets in (or doesn't, in which case it would need to be implemented here instead of plot options support dropped for now and deferred).

PR #16, opened 2023-05-12
prevent closing the "flux-vs-time" viewer by injecting css into app.vue

kecnry:
This PR implements a barebones widget with no elements but allows injecting the style into the app-level in jdaviz, and is used here to hide the close button on the "flux-vs-time" viewer tab in golden layout.

NOTE: this requires [unreleased but merged changes upstream](https://github.com/spacetelescope/jdaviz/pull/2200) and ~the "flux-vs-time" viewer doesn't actually exist in main here yet~, but just wanted this to be reviewed on its own rather than get buried in with the upcoming phase-viewer PR.

~To test, feel free to test on top of #15 or edit the style to change the background color (or something).  Otherwise we can wait until #15 is merged and rebase on top of that.~

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -3,6 +3,7 @@ import astropy.units as u
 from lightkurve import LightCurve
 
 from jdaviz.core.helpers import ConfigHelper
+from lcviz.plugins import StyleWidget
 
 __all__ = ['LCviz']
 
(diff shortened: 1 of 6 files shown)


kecnry:
Merging here without approval since the upstream PR was approved (and included this in that review).

PR #19, opened 2023-05-17
phase viewer and ephemeris plugin

kecnry:
This PR introduces a new ephemeris plugin with support for multiple ephemeris "components", which then creates and updates phase arrays and viewers which are kept up-to-date with the ephemeris and linked correctly to support subsets.   This also includes a very simple interface to periodagrams which just propose the top peak - note that we will likely split periodograms/FFTs/PDM into their own plugins which interact with the ephemeris plugin, but this is at least useful for now. 

https://github.com/spacetelescope/lcviz/assets/877591/0a765a63-b23d-442d-b7b9-0064d3c18579

Possibly in scope:
* components sometimes fail to be added/renamed
* viewers occasionally crash (likely upstream bugs)

Intentionally deferred:
* full periodogram support (input arguments, plot, interactivity)
* ability to duplicate phase-viewers
* aliasing support in phase-viewers
* reference time incorrect when loading multiple light curves
* prevent re-ordering of viewers when changing component name


rosteen:
Starting to test this, my initial reaction is that I like it other than one minor quibble and a possible bug. The quibble is that I'm not sure I like having the viewer automatically appear the first time you interact with the plugin. It seems a little arbitrary that I change, for example, the period, and suddenly the viewer appears. I wonder if it would be better to either have the default phase folded viewer be tabbed but unpopulated on load, or have a button to create the viewer.

The bug is that when the viewer first appears, the data isn't folded correctly (or at least isn't displaying correctly), and then fixes itself when I tweak any of the values slightly. In the screenshots below I first set the period to 5, and
 then the second screenshot is after clicking the arrow to increment it up by .001:

Screenshot 2023-05-22 at 5 25 11 PM


Screenshot 2023-05-22 at 5 25 28 PM

rosteen:
This is for the example data, it's not an aliasing issue since it doesn't depend on what the initial period you put in, here you can see it happens if you initialize to either the Lomb-Scargle calculated value or to a user input. I forgot to show this, but if you go from the initial value of 5 to 5.001 and then back to 5, the plot looks fine with period of 5 the second time.

https://github.com/spacetelescope/lcviz/assets/39831871/918a6a45-cb21-4260-bf72-3ee234a84d04

kecnry:
That is bizarre - I can't reproduce on my end.  If you call the following when the phasing is incorrect (before updating anything to fix it otherwise), does it show the correct ephemeris and then update the plot correctly?

```
ephem = lcviz.plugins['Ephemeris']
print(ephem.ephemerides)
ephem._obj._update_all_phase_arrays()
```

rosteen:
It prints the correct value but doesn't update the plot:

Screenshot 2023-05-23 at 11 01 50 AM

I update my glue-jupyter and bqplot installations just in case, didn't help. Maybe I should update glue-core as well.

kecnry:
ok, then can I ask you to just try to parse `_update_all_phase_arrays`, add some debug print statements, and try to track down why it isn't actually updating the array (or maybe setting the links correctly)? 😬   The method _should_ take the ephemeris (which is updated) and update the plot.  It's very odd that it does once making another change 🤔

lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -0,0 +1,383 @@
...
+
+        # update value in the dictionary (to support multi-ephems)
+        if event:
+            self.update_ephemeris(**{event.get('name'): event.get('new')})
+            # will call _update_all_phase_arrays
+        else:
+            self._update_all_phase_arrays(component=self.component_selected)
+
+        # if phase-viewer doesn't yet exist in the app, create it now
+        self.create_phase_viewer()
```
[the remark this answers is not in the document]
kecnry: wow, that surprises me... if anything I'd expect this to just need an extra viewer-update, but maybe that's what is causing the fix.  Thanks for tracking it down!

rosteen (approved):
I didn't go over the code with too much rigor, but after the fix to the initial viewer population everything seems to work pretty well. The period finding doesn't seem very good, but I think that's an consequence of the example light curve being a mess with lots of stellar variability. We can definitely iterate from here.

bmorris3:
Screen Shot 2023-05-24 at 14 39 38
Are you expecting the warning shown here? (Using jdaviz on main.)

kecnry:
jdaviz main will now require https://github.com/spacetelescope/lcviz/pull/20 (or just use jdaviz before https://github.com/spacetelescope/jdaviz/pull/2199)

bmorris3:
Just had to share – this works so beautifully for use cases on spotted, flaring stars. Thanks to @jradavenport for the rotation period measurement 👌🏻 .

Screen Shot 2023-05-24 at 15 01 42

lcviz/plugins/ephemeris/ephemeris.vue
```diff
@@ -0,0 +1,137 @@
...
+    </v-row>
+
+    <div v-if="component_items.length > 0">
+      <j-plugin-section-header>Ephemeris</j-plugin-section-header>
+
+      <v-row>
+        <v-text-field
+          ref="t0"
+          type="number"
+          label="Zeropoint (t0)"
```
bmorris3, line 30: Not to nitpick too much, but I think "Zeropoint" is too vague. If t0 always refers to the reference time in the light curve over time (viewer on the left), then maybe `t0` should be "Reference time" or "Time at zero phase".

lcviz/plugins/ephemeris/ephemeris.vue
```diff
@@ -0,0 +1,137 @@
...
+
+    <lcviz-editable-select
+      :mode.sync="component_mode"
+      :edit_value.sync="component_edit_value"
+      :items="component_items"
+      :selected.sync="component_selected"
+      label="Component"
+      hint="Select an ephemeris component."
+      >
+    </lcviz-editable-select>
```
bmorris3, line 15: It isn't clear to me what this is meant to do, which might mean we need a better name/label. It seems like this controls _part of_ the name of the ephemeris tab in the viewer, which is the viewer reference name. 

I think you named it "Component" because it could be a "component" of the physical system, like planet b or c, or the rotation period of a primary or secondary "component" of a binary. Is that right? If so, maybe a better name for this box is "Ephemeris viewer label" or similar.

lcviz/plugins/ephemeris/ephemeris.vue
```diff
@@ -0,0 +1,137 @@
...
+          persistent-hint
+          :rules="[() => period!=='' || 'This field is required',
+                   () => period > 0 || 'Period must be greater than zero']"
+        >
+          <template v-slot:append>
+            <j-tooltip tooltipcontent="halve period">
+              <v-icon style="cursor: pointer" @click="period_halve">mdi-division</v-icon>
+            </j-tooltip>
+            <j-tooltip tooltipcontent="double period">
+              <v-icon style="cursor: pointer" @click="period_double">mdi-multiplication</v-icon>
```
bmorris3, line 57: Just wanted to add here that I had expressed early on that I might want something other than the mult/div buttons, but now that I've played with it this way, it's actually really clear that this is a handy and intuitive interface. Neat!

lcviz/plugins/ephemeris/ephemeris.vue
```diff
@@ -0,0 +1,137 @@
...
+            </j-tooltip>
+          </template>
+        </v-text-field>
+      </v-row>
+
+      <v-row>
+        <v-text-field
+          ref="t0"
+          type="number"
+          label="Period derivative (dpdt)"
```
bmorris3, line 67: Do we want to expose these variable names in the labels (`dpdt`)?

bmorris3 (approved):
This looks great. Feel free to merge after you've responded to the comments below (even if you don't adopt all suggestions; I just want to start these conversations).

lcviz/plugins/ephemeris/ephemeris.vue
kecnry: I struggled with this a bit as well, and think as we start to write other plugins it might become more clear and we might want to revise this.  Elsewhere I think we'll want to refer to this label just as "ephemeris" - `binning_plugin.ephemeris` to bin the data phased by a specific ephemeris, for example - but that feels awkward here since I wanted to use the `ephemeris` property to return the actual ephemeris information.  I do want to keep some synergy between the plugin attribute name (currently `ephemeris_plugin.component`) and the UI.

bmorris3: Understood. The crux of my concern is that the word "component" has too many meanings (glue Component? system component? viewer component?).

kecnry: agreed (and especially with model components), please do bring this back up when we get more context with other plugins starting to fall into place and let's see if we can come up with a consistent but intuitive solution.

PR #22, opened 2023-05-26
enable coords-info and markers plugin

kecnry:
This PR implements and enables both mouseover coordinates and the markers plugin.  Unlike the implementation in specviz, the neareset-point logic is done based on the nearest point in the pixel-space, so that highlighting points in eclipses/transits feels more natural.

https://github.com/spacetelescope/lcviz/assets/877591/257f0cda-666b-4379-b3fa-7e16e2a1b53b


duytnguyendtn:
Basic functionality looks good! Though I'm running into a bug and I can't tell if its' with this PR or with the ephemeris/phase viewer. I have not adjusted any parameters in the plugin, just showed the viewer:

image

If I kick the period to something else and force the points to redraw, the hoverover seems to be constrained properly

kecnry:
Ok, can you file an issue for that (I personally don't think it should hold up this PR)?  @rosteen ran into [something similar in the review of the phase-folding plugin](https://github.com/spacetelescope/lcviz/pull/19#issuecomment-1558044055), but apparently the fix didn't handle every situation.  I'm guessing its actually an upstream bug, but I cannot reproduce it on my machine and it will need more investigation.

duytnguyendtn (approved):
Functionality works well (with the exception for the above bug) and the code looks as I would expect. Thanks!

@bmorris3 I feel like this PR is straight forward enough. You can decide whether you'd like to still review!

PR #23, opened 2023-05-30
flatten plugin

kecnry:
This PR implements a wrapper around the `flatten` method from `lightkurve`, with live-preview of the trend as well as the detrended light curve in time-space and any phase-folded viewers.

https://github.com/spacetelescope/lcviz/assets/877591/5174af32-690e-4886-bd48-70b41a1b8892

TODO:
* fix case where not overwriting results in a traceback

Future follow-up:
* option to export trend (a la residuals in model fitting)
* ability to maintain zoom limits and plot options when replacing data (would need to be done in `jdaviz`)


bmorris3:
This is kind of a meta comment, but is there a way to make plugin implementations for exposing lightkurve functions smaller somehow? There are [quite a few methods](https://docs.lightkurve.org/reference/lightcurve.html), so I'm wondering how many more +400 PRs it'll take to cover the ones we decide that "we need."

lcviz/plugins/flatten/flatten.py
```diff
@@ -0,0 +1,200 @@
...
+    show_live_preview = Bool(True).tag(sync=True)
+    default_to_overwrite = Bool(True).tag(sync=True)
+    flatten_err = Unicode().tag(sync=True)
+
+    window_length = IntHandleEmpty(101).tag(sync=True)
+    polyorder = IntHandleEmpty(2).tag(sync=True)
+    break_tolerance = IntHandleEmpty(5).tag(sync=True)
+    niters = IntHandleEmpty(3).tag(sync=True)
+    sigma = FloatHandleEmpty(3).tag(sync=True)
+    unnormalize = Bool(False).tag(sync=True)
```
bmorris3, line 51: I'm a bit conflicted about default False for unnormalize. I appreciate the descriptions and warning messages in the plugin, but I think it's still surprising to show a live preview on top of the rate LC when clicking the "FLATTEN" button produces a normalized light curve. 

Normalization is a common and non-unique task that we may actually need control over in its own plugin, ultimately. Normalization can be done with different stats for different science (median/mean/min/max), and can use a zero or one-centered convention depending on what's convenient for your analysis.

In particular, I'm thinking about this workflow. Say you download a Kepler LC as an electron rate time series, load into LCviz, estimate an EB/transiting planet orbital period and phase fold with Ephemeris, flatten, and fit with a simple eclipse model. Depending on the model's developer and the mass of the secondary (!), the convention changes about whether to normalize to zero (common for ~200 ppm Earth-sized transits) or one (truly relative photometry). We could homogenize the inputs and outputs to the models, but part of the joy (/curse) of photometry is that everyone has a model, so that seems like a lot of work. And some components of photometric models are formulated for addition in relative flux units (like adding a phase curve or ellipsoidal variations to a transit model), while others are multiplicative (like multiplying an exoplanet secondary eclipse to a phase curve model). You may have to carefully unnormalize and renormalize every time you enter or exit interactions with those models.

Plus, the units have to be handled differently – your flattened data are count rates if `unnormalize=True` and dimensionless ratios otherwise. We'll have to keep track if/how the light curve has been normalized somewhere ultimately, so perhaps a dedicated plugin is the right approach?

I'll stop ranting, thanks for coming to my TED talk 😄

lcviz/tests/test_plugin_flatten.py
```diff
@@ -0,0 +1,56 @@
...
+    assert f._obj.add_results.label != orig_label
+    f.flatten(add_data=True)
+
+
+def test_unnormalize(helper, light_curve_like_kepler_quarter):
+    helper.load_data(light_curve_like_kepler_quarter)
+
+    f = helper.plugins['Flatten']
+
+    assert f.unnormalize is False
```
bmorris3, line 49:
```suggestion
    assert not f.unnormalize
```

lcviz/tests/test_plugin_flatten.py
```diff
@@ -0,0 +1,56 @@
...
+    helper.load_data(light_curve_like_kepler_quarter)
+
+    f = helper.plugins['Flatten']
+
+    orig_label = f.dataset.selected
+    assert f._obj.add_results.label_overwrite is True
+    assert f._obj.add_results.label == orig_label
+    f.default_to_overwrite = False
+    assert f._obj.add_results.label_overwrite is False
+    assert f._obj.add_results.label != orig_label
```
bmorris3, line 40:
```suggestion
    assert f._obj.add_results.label_overwrite
    assert f._obj.add_results.label == orig_label
    f.default_to_overwrite = False
    assert not f._obj.add_results.label_overwrite
    assert f._obj.add_results.label != orig_label
```

lcviz/tests/test_plugin_flatten.py
```diff
@@ -0,0 +1,56 @@
...
+    f.open_in_tray()
+    ephem = helper.plugins['Ephemeris']
+    pv = ephem.create_phase_viewer()
+
+    assert len(_get_marks_from_viewer(tv)) == 2
+    assert len(_get_marks_from_viewer(pv)) == 1
+
+    orig_label = f.dataset.selected
+    assert f.dataset.selected_obj is not None
+    assert f._obj.add_results.label_overwrite is True
```
bmorris3, line 25:
```suggestion
    assert f._obj.add_results.label_overwrite
```

bmorris3 (commented):
Some questions from this demo (using jdaviz on main): 

https://github.com/spacetelescope/lcviz/assets/3497584/555a7d95-e0ed-4c95-b6c9-9f3b41069674

The live-updating, blue, flattened LC looks great. Then when you click "flatten": 
1. something is flashed to the flux-vs-time viewer, but I'm not sure what it is
2. the result is plotted in blue as though it's still a marker, but now it's the only dataset in the viewer (right?). If flatten has overwritten the original dataset, should we preserve the default LC styling since it's not "new" data nor a live updating preview?
3. Should "Show In Viewer" be visible in flatten? It doesn't seem like I have the option.

I'd advocate for adding the flattened light curve to the data collection and giving the user the option to remove the original data from the viewer within the flatten plugin, but not overwriting by default. It's possible to do irreversible detrending to a light curve, then realize you shouldn't have, and then have to start from scratch because you've overwritten the original.

kecnry:
> something is flashed to the flux-vs-time viewer, but I'm not sure what it is

This is the viewer emptying and then the new attributes being assigned (more on this in the response to your next comment)

> the result is plotted in blue as though it's still a marker, but now it's the only dataset in the viewer (right?). If flatten has overwritten the original dataset, should we preserve the default LC styling since it's not "new" data nor a live updating preview?

Yes, I totally agree and had this written down as something I wanted to do upstream.  By replacing the data components instead of just loading a new data with the same label, this should also fix the flashing annoyance.  Let me make sure there is a ticket for that... but since it is a change in behavior in jdaviz, I don't think its worth overriding here... we should just "fix" it there)

> Should "Show In Viewer" be visible in flatten? It doesn't seem like I have the option.

You don't have the option if the label is overwriting.  If you override the label to something custom, then it becomes enabled (again, upstream behavior).  Having the ability to immediately show in _multiple_ viewers is also something that this can motivate adding upstream.

> I'd advocate for adding the flattened light curve to the data collection and giving the user the option to remove the original data from the viewer within the flatten plugin, but not overwriting by default. It's possible to do irreversible detrending to a light curve, then realize you shouldn't have, and then have to start from scratch because you've overwritten the original.

Would just defaulting `default_to_overwrite` (one of the switches in the plugin "settings") achieve what you'd expect?

bmorris3:
> Would just defaulting default_to_overwrite (one of the switches in the plugin "settings") achieve what you'd expect?

For defaults, I'm thinking: `default_to_overwrite = False`, plus a new `hide_original = True` (removes corresponding data entry from the viewer).

lcviz/plugins/flatten/flatten.py
kecnry: I was/am conflicted as well.  The reason I proposed this was that: I think its super useful to see the live-preview, the live-preview is only actually useful if it can be seen on the axes limits, and by default it seems weird to get something different than the live preview.  Additionally, if you don't overwrite the input data, then it can be jarring (although it makes sense) to have to reset the limits to a completely different range.  For "first-look" type workflows, I think "unnormalizing" is just more convenient, and so made that the default.

But totally agreed it needs to be an option as there are different use-cases and I'm all in favor of a more advanced (probably separate) normalization plugin down the road.

With that said - do you still think we should change the default?  (Either way, changing the default of a switch is super easy to do later, as long as we're in agreement on how the switch itself behaves for now).

kecnry:
> plus a new hide_original = True (removes corresponding data entry from the viewer).

I'll have to think about that... my gut instinct says that logic should belong upstream in the shared "add results" component, but I'm not sure if there are other situations where it would be useful or not.

lcviz/plugins/flatten/flatten.py
bmorris3: Understood. I agree that "unnormalize=False" is the correct default so that you see the flattened light curve overplotted on the original data. But I might go so far as to suggest we keep the attribute but don't show the switch in the plugin, so we don't have to remove it later?

PR #24, opened 2023-05-31
Support full round-trip for LightCurve <-> Data

bmorris3:
Before this PR, the glue translator for `LightCurve` only translated a few columns from the underlying astropy `TimeSeries` object to Data Components, and vice versa. 

This PR supports translation of all columns in a Kepler light curve (and hopefully other missions too). To test the feature, I've introduced a FITS file with a segment of a Kepler light curve that contains only 10 rows/times, but all columns in a real Kepler light curve retrieved from MAST (40 kb). 

I still rely on lightkurve to parse the FITS file, rather than passing the path to `lcviz.load_data`. This is a workaround to avoid downloading a FITS file with lightkurve, since https://github.com/lightkurve/lightkurve/pull/1299 has not yet been released (as first noted in https://github.com/spacetelescope/lcviz/pull/10).

This translator may need updates in the future if we add Components that should not be translated into columns in the `TimeSeries`.


kecnry:
this seems to result in the `flux-vs-time` viewer showing `time` vs `dt` instead of... `flux` vs `time` (although now in main I'm seeing `uncertainty` vs `flux` which is also wrong, so maybe its not entirely related to this PR?) 🤔

bmorris3:
> maybe its not entirely related to this PR?

I think that's unrelated – I'm seeing the correct axes on this branch.
Screen Shot 2023-05-31 at 13 31 21

kecnry:
ok, I am seeing the correct behavior with latest on jdaviz main, I think the display units branch is causing issues, but we'll track that there.  Sorry for the noise!

kecnry (approved):
Looks good to me and can confirm this is fixing the crashing issue I was seeing when attempting to load results from `get_object().bin()` back into the app.

PR #28, opened 2023-06-13
binning plugin

kecnry:
This PR implements a binning plugin (wrapping existing `lightkurve` binning functionality), supporting binning in either time or phase-space, along with a live-preview that works in conjunction with changes to the underlying ephemeris (when phase-binning, the binning is recomputed with any change to the ephemeris that dictates that phase-folding).  Once loaded into the app, the binned data remain fixed (and with https://github.com/spacetelescope/jdaviz/pull/2312 would only be available to load in the respective viewer).

This PR currently only exposes the number of bins as an input parameter, but we might want to consider (now or as a follow-up effort) the ability to optionally provide a bin width in time/phase units instead.

Support for having live-updating binning in the data-collection object might be considered in the future, but is out-of-scope for now.

https://github.com/spacetelescope/lcviz/assets/877591/b183c3bb-9c31-4e9e-8125-e8111e800c8c

~There seems to be a slight bug with the reference time when the time-binned data is actually applied and then phased (the phasing in the preview and the finalized data appears different).  I'm not sure if that is introduced here or an existing bug that this uncovers.~

The change in color when adding the time-binned data to the second viewer is likely an upstream bug.

* [plugin docs](https://lcviz--28.org.readthedocs.build/en/28/plugins.html#binning)
* [plugin API docs](https://lcviz--28.org.readthedocs.build/en/28/api/lcviz.plugins.binning.binning.Binning.html)

### DIFF
--- a/lcviz/plugins/binning/binning.py
+++ b/lcviz/plugins/binning/binning.py
@@ -12,6 +12,7 @@ from jdaviz.core.user_api import PluginUserApi
 from lcviz.events import EphemerisChangedMessage
 from lcviz.helper import _default_time_viewer_reference_name
 from lcviz.marks import LivePreviewBinning
+from lcviz.parsers import _data_with_reftime
 from lcviz.template_mixin import EphemerisSelectMixin
 
 
@@ -196,7 +197,8 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
             self.app._jdaviz_helper._set_data_component(data, phase_comp_lbl, lc.time.value)
 
         else:
-            data = None
+            # need to send through parser-logic to assign the correct reference time
+            data = _data_with_reftime(self.app, lc)
 
         if add_data:
             # add data to the collection/viewer
--- a/lcviz/plugins/binning/binning.py
+++ b/lcviz/plugins/binning/binning.py
@@ -42,7 +42,7 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
     show_live_preview = Bool(True).tag(sync=True)
 
     n_bins = IntHandleEmpty(100).tag(sync=True)
-    map_to_times = Bool(True).tag(sync=True)
+    map_to_times = Bool(False).tag(sync=True)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -186,6 +186,7 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
 
         lc = input_lc.bin(time_bin_size=(input_lc.time[-1]-input_lc.time[0]).value/self.n_bins)
         if self.ephemeris_selected != 'No ephemeris':
+            # lc.time.value are actually phases, so convert to times starting at time t0
             times = self.ephemeris_plugin.phases_to_times(lc.time.value, self.ephemeris_selected)
 
             if self.map_to_times:
@@ -193,19 +194,24 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
                 binned_lc = lc.copy()
                 t0 = self.ephemeris_dict.get('t0', 0.0)
                 period = self.ephemeris_dict.get('period', 1.0)
-                lc.time = times
+                lc.time = Time(times,
+                               format=input_lc.time.format,
+                               scale=input_lc.time.scale)
 
                 # extend forward and backwards in cycles for the full range of input_lc
                 min_time, max_time = input_lc.time_original.min().value, input_lc.time_original.max().value
                 for start_time in np.arange(min_time, max_time, period):
                     this_times = start_time + (times - t0)
                     if start_time == min_time:
-                        lc.time = this_times
+                        lc.time = Time(this_times,
+                                       format=input_lc.time.format,
+                                       scale=input_lc.time.scale)
                     else:
                         binned_lc.time = this_times
                         lc = lc.append(binned_lc)
 
             else:
+                # then just set the time_original column, leaving the time column as phases
                 time_col = Time(times,
                                 format=input_lc.time_original.format,
                                 scale=input_lc.time_original.scale)
@@ -213,6 +219,7 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
 
         if add_data:
             # add data to the collection/viewer
+            # NOTE: lc will have _LCVIZ_EPHEMERIS set if phase-folded
             self.add_results.add_results_from_plugin(lc)
 
         return lc
--- a/lcviz/plugins/binning/binning.py
+++ b/lcviz/plugins/binning/binning.py
@@ -1,0 +1,156 @@
+from traitlets import Bool, observe
+
+from jdaviz.core.custom_traitlets import IntHandleEmpty
+from jdaviz.core.events import (ViewerAddedMessage, ViewerRemovedMessage)
+from jdaviz.core.registries import tray_registry
+from jdaviz.core.template_mixin import (PluginTemplateMixin,
+                                        DatasetSelectMixin, AddResultsMixin)
+from jdaviz.core.user_api import PluginUserApi
+
+from lcviz.helper import _default_time_viewer_reference_name
+from lcviz.marks import LivePreviewBinning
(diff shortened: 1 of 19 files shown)


bmorris3:
By default, the binned results have the same marker properties as the default (except an increment in the color cycle). This makes sense most of the time, but in the specific case of binning, it's useful to have the binned markers larger and/or with a different marker shape, since it can be hard to see the bins over the busy original, densely sampled data:

bqplot (23)

Is there a way to specify the marker size/shape for binned results specifically?

lcviz/plugins/binning/binning.vue
```diff
@@ -0,0 +1,72 @@
...
+      hint="Select the light curve as input."
+    />
+
+    <plugin-ephemeris-select
+      :items="ephemeris_items"
+      :selected.sync="ephemeris_selected"
+      :show_if_single_entry="false"
+      label="Ephemeris"
+      hint="Select the phase-folding as input."
+    />
```
bmorris3, line 41: It's clear that using the "ephemeris selected" machinery is the correct choice here, but I think the naming is a bit confusing if you haven't yet specified an ephemeris. For example, if I load a light curve at 1 min cadence but I want to see what happens when binned on ~year timescales, it'll be unclear what "ephemeris" means, since I don't have a strict period or epoch in mind. 

My instinct was that this selector could be a "viewer" selector, so that one option says "flux-vs-time". What do you think?

lcviz/plugins/binning/binning.py
```diff
@@ -0,0 +1,194 @@
...
+      Label of the component corresponding to the active ephemeris.
+    * :meth:`input_lc`
+      Data used as input to binning, based on ``dataset`` and ``ephemeris``.
+    * ``n_bins``
+    * ``add_results`` (:class:`~jdaviz.core.template_mixin.AddResults`)
+    * :meth:`bin`
+    """
+    template_file = __file__, "binning.vue"
+
+    show_live_preview = Bool(True).tag(sync=True)
```
bmorris3, line 40: I see that the default is `True`, but it looks like the preview doesn't get displayed. I was able to get a live preview after clicking `Bin`, but not before:

Screen Shot 2023-07-31 at 14 09 23

kecnry:
> Is there a way to specify the marker size/shape for binned results specifically?

Sure, we can do that with layer defaults (see latest commit).  If we want that to be more advanced (dependent on the number of bins, etc), then we'll either have to store more information in the metadata, or add support at the `add_results_to_plugin` level (but that would be forgotten when the data is removed and re-added to the viewer).

lcviz/plugins/binning/binning.vue
```diff
@@ -0,0 +1,72 @@
...
+      :label_auto.sync="results_label_auto"
+      :label_invalid_msg="results_label_invalid_msg"
+      :label_overwrite="results_label_overwrite"
+      label_hint="Label for the binned data."
+      :add_to_viewer_items="add_to_viewer_items"
+      :add_to_viewer_selected.sync="add_to_viewer_selected"
+      action_label="Bin"
+      action_tooltip="Bin data"
+      @click:action="apply"
+    ></plugin-add-results>
```
bmorris3, line 69: Should we track a progress status with a spinner here, like in Imviz:

https://github.com/spacetelescope/jdaviz/blob/b391423294fb5e6cd5ef6d7cafbb7a0b05a66ca3/jdaviz/configs/imviz/plugins/links_control/links_control.vue#L65-L79

This is a bit slow for me with large datasets and the data looks strangely linear at first as it loads. Graying out the viewer during load would help there.

lcviz/plugins/binning/binning.vue
```diff
@@ -0,0 +1,72 @@
...
+        label="N Bins"
+        type="number"
+        v-model.number="n_bins"
+        :step="10"
+        :rules="[() => n_bins !== '' || 'This field is required',
+                 () => n_bins > 0 || 'Number of bins must be positive']"
+        hint="Number of bins."
+        persistent-hint
+      >
+      </v-text-field>
```
bmorris3, line 54: `N bins` is a good start, but it would be awesome/helpful to also offer a bin width in time units (10 minute bins, 5 day bins, etc), and do the arithmetic for the user. Your choice if that's in this PR or not.

lcviz/plugins/binning/binning.py
```diff
@@ -0,0 +1,194 @@
...
+            if mark.visible:
+                mark.clear()
+                mark.visible = False
+
+    @observe("dataset_selected", "ephemeris_selected")
+    def _set_default_results_label(self, event={}):
+        '''Generate a label and set the results field to that value'''
+        if not hasattr(self, 'ephemeris'):
+            return
+        label = f"binned {self.dataset_selected}"
```
bmorris3, line 104: This is ok for now, but it'd be helpful to add the number of bins in this string too. For example, here I've binned data in each viewer at different bin numbers in different data entries:

Screen Shot 2023-07-31 at 14 23 34

but the data labels in the drop down and in the data collection only show:

```
DataCollection (3 data sets)
	  0: KIC 10748390[pdcsap_flux]
	  1: binned KIC 10748390[pdcsap_flux]:default
	  2: binned KIC 10748390[pdcsap_flux]
```
You and I know that `default` means an ephemeris is being applied to `1`, but that might not obvious for the user. We could do `f'binned (N={self.nbins}) {self.dataset_selected}'` as a start.

lcviz/plugins/binning/binning.py
```diff
@@ -0,0 +1,194 @@
...
+        if self.ephemeris_selected not in self.ephemeris._manual_options:
+            label += f":{self.ephemeris_selected}"
```
bmorris3, line 106: And maybe this could be more like `f", Ephemeris: {self.ephemeris_selected}"` so it's clear what `default` refers to.

lcviz/plugins/binning/binning.py
kecnry: this is because of changes in jdaviz 3.6 - fix coming soon once #33 gets in!

lcviz/plugins/binning/binning.vue
kecnry: Agreed!  This gets more complicated when switching between time and phase-binning, so my vote would be to defer this until we get a better feel for what would feel intuitive for these cases

lcviz/plugins/binning/binning.vue
kecnry: this shouldn't even show in the UI if you haven't manually created an ephemeris, so I'm hoping it'll make more sense in that context.  But this is definitely an unsolved problem from the API that might warrant some thinking 🤔

lcviz/plugins/binning/binning.py
kecnry: (should be fixed now)

lcviz/plugins/binning/binning.vue
kecnry: If the live-preview is enabled, then we can probably skip rebinning when clicking the button and it should be quite responsive (just the time to load into the data collection/viewer).  But in that case we might want some sort of indication that the binning is updating in the preview.  And if the live-preview is off, then an indication when clicking the button will still be necessary.

This same pattern should probably also apply to flattening 🤔

lcviz/plugins/binning/binning.vue
kecnry: for the API, something like https://github.com/spacetelescope/jdaviz/pull/2347 would help (but that hasn't seemed to get much traction/support yet).

lcviz/plugins/binning/binning.py
kecnry: I'm going to open a ticket ([:cat:](https://jira.stsci.edu/browse/JDAT-3789)) to consider this in general across all plugins that export data back to the app.  I think we need a consistent and somewhat comprehensive approach for (1) what information is most important to be in the default labels, and (2) a consistent "syntax" (for example, the ephemeris in the phase-viewer labels is consistent with the `:default` suffix right now, but we might want to consider changing that too).

Of course we probably can't cover every use-case without undesirably long labels, which is why the labels can be manually modified.

A really cool idea that we could implement upstream would be the concept of user-defined templates with some sort of drag and drop UI, but I'm not sure that's worth the effort and overhead.  Or maybe any inputs that have been "touched" are included, but others that have remained fixed aren't?  Or maybe we can store all the inputs in the metadata and have dynamic labels after the fact?

bmorris3 (approved):
Let's get this in! 🚀

PR #30, opened 2023-06-15
basic frequency analysis plugin

kecnry:
This PR implements a frequency/period analysis plugin that displays and provides API access to BLS and Lomb-Scargle periodograms, in either frequency or period space, and with automatic or a manual range of input frequencies/periods.

Note that this requires the dev version of jdaviz for the [plugin plot subcomponent](https://github.com/spacetelescope/jdaviz/pull/2254).

Potential follow-up efforts:
- add toolbar to plugin plot subcomponents with zoom/pan tools
- support additional input arguments per-method
- support optional interactive setting of frequency/period limits (via zoom limits on flux-vs-time viewer)
- report peaks and share code back with ephemeris plugin (which currently reports the max peak, but without the ability to set any input options)
- support interactive click & drag to update selected ephemeris


lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -0,0 +1,158 @@
...
+    method_items = List().tag(sync=True)
+    method_selected = Unicode().tag(sync=True)
+
+    xunit_items = List().tag(sync=True)
+    xunit_selected = Unicode().tag(sync=True)
+
+    auto_range = Bool(True).tag(sync=True)
+    minimum = FloatHandleEmpty(1).tag(sync=True)  # period and immediately converted to freq
+    minimum_step = Float(0.1).tag(sync=True)
+    maximum = FloatHandleEmpty(10).tag(sync=True)  # period and immediately converted to freq
```
bmorris3, line 46: Loading the Q10 Kepler LC of HAT-P-11, the new plugin starts with min=1, max=10, irrespective of the dataset selected. For example, if I do this:
```python
fa = lcviz.plugins['Frequency Analysis']
fa.xunit = 'period'
fa.auto_range = True
fa.dataset.selected_obj = lcviz.app.data_collection[0].label
fa.maximum, fa.minimum
```
I see 10 and 1. 

Can we have this update on data selection, based on the timespan of the selected dataset? For most datasets, it would be helpful to offer an autorange between (a few)*(exptime) and (obs duration)/(a few).

The above min/max attrs are then confusing because with `auto_range` selected, the periodogram limits are 0 and ~450:
Screen Shot 2023-07-07 at 09 14 53

Any ideas what's going on there?

lcviz/plugins/frequency_analysis/frequency_analysis.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <j-tray-plugin
+    description='Periodograms and FFTs.'
```
bmorris3, line 3: Should we mention FFTs before knowing we're going to support them?

lcviz/plugins/frequency_analysis/frequency_analysis.vue
```diff
@@ -0,0 +1,110 @@
...
+    />
+
+    <v-row>
+      <v-select
+        :menu-props="{ left: true }"
+        attach
+        :items="method_items.map(i => i.label)"
+        v-model="method_selected"
+        label="Algorithm/Method"
+        hint="Method to determine period."
```
bmorris3, line 22:
```suggestion
        hint="Method to determine power at each frequency or period."
```

lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -0,0 +1,158 @@
...
+            except Exception as err:
+                self.spinner = False
+                self.err = str(err)
+                self.plot.clear_all_marks()
+                return None
+        elif self.method == 'Lomb-Scargle':
+            try:
+                per = periodogram.LombScarglePeriodogram.from_lightcurve(self.dataset.selected_obj,
+                                                                         minimum_period=min_period,
+                                                                         maximum_period=max_period)
```
bmorris3, line 114: Some [options that we don't yet offer](https://docs.lightkurve.org/reference/api/lightkurve.periodogram.LombScarglePeriodogram.from_lightcurve.html#lightkurve.periodogram.LombScarglePeriodogram.from_lightcurve) to consider adding in the future include `nterms` and `normalization`, probably in that order of priority.

lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -0,0 +1,158 @@
...
+                self.spinner = False
+                self.err = str(err)
+                self.plot.clear_all_marks()
+                return None
+        else:
+            self.spinner = False
+            raise NotImplementedError(f"periodogram not implemented for {self.method}")
+
+        self.spinner = False
+        return per
```
bmorris3, line 125: In the Ephemeris plugin, we have a Period Finding feature which suggests a period based on the peaks in the periodogram.
image
With the addition of Frequency Analysis, this section of the ephemeris is redundant and kind of ambiguous – what are the options you could give to the Frequency Analysis plugin that reproduces the guess in the Ephemeris plugin?

Maybe we should turn off that part of the Ephemeris plugin, and expect the user to use Frequency Analysis instead?

lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -0,0 +1,158 @@
...
+            min_period, max_period = None, None
+        elif self.xunit_selected == 'period':
+            min_period, max_period = self.minimum, self.maximum
+        else:
+            min_period, max_period = self.maximum ** -1, self.minimum ** -1
+        if self.method == 'Box Least Squares':
+            try:
+                per = periodogram.BoxLeastSquaresPeriodogram.from_lightcurve(self.dataset.selected_obj,  # noqa
+                                                                             minimum_period=min_period,  # noqa
+                                                                             maximum_period=max_period)  # noqa
```
bmorris3, line 104: L-S is convenient because it gives ~"one answer" per light curve, whereas BLS has additional arguments with produce important differences in the result. The critical [missing "optional" parameter](https://docs.lightkurve.org/reference/api/lightkurve.periodogram.BoxLeastSquaresPeriodogram.from_lightcurve.html) in this implementation is `duration`. BLS does a least squares fit with a box model _of a specific duration_. By default this BLSP method in lightkurve computes the periodogram with a grid of durations ([0.05, 0.10, 0.15, 0.20, 0.25, 0.33] d). The HAT-P-11 example works well because the duration is close to 0.1 d. If the box duration does not match the transit you're trying to fit, you won't find the transit, so we really need to add a duration setter box for the BLS method.

lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # convert minimum/maximum for next computation
+        self._ignore_auto_update = True
+        old_min, old_max = self.minimum, self.maximum
+        self.minimum = old_max ** -1 if old_max > 0 else 0
+        self.maximum = old_min ** -1 if old_min > 0 else 0
+        self._ignore_auto_update = False
+
+    @observe('dataset_selected', 'method_selected', 'auto_range', 'minimum', 'maximum')
+    def _update_periodogram(self, *args):
```
bmorris3, line 144: Is this method working as intended? If I make a periodogram with the HAT-P-11 dataset, use the Flatten plugin to remove stellar rotation, and then return to the Frequency Analysis plugin, the periodogram disappears:
image
I haven't figured out how to make it reappear.

bmorris3 (changes requested):
Lots of notes. This is fun 😄

lcviz/plugins/frequency_analysis/frequency_analysis.py
bmorris3: Also, what are the units on these axes?

lcviz/plugins/frequency_analysis/frequency_analysis.py
bmorris3: I know that _we_ know this, but should we put a note in the Frequency Analysis plugin that indicates that it operates on the data in the `flux-vs-time` viewer?

kecnry: > Maybe we should turn off that part of the Ephemeris plugin, and expect the user to use Frequency Analysis instead?

As we address follow-up tickets and add more flexibility (being able to click on the plot to set the ephemeris), we may very well decide to remove the functionality from the ephemeris plugin, or at the least, share more code between the two.  But I think for now it is useful to have (I use it all the time to quickly phase).

> I know that we know this, but should we put a note in the Frequency Analysis plugin that indicates that it operates on the data in the flux-vs-time viewer?

There will already be a data dropdown if more than one data entry exists.  I think its self-explanatory that it acts on flux vs time instead of phase (or uncertainty vs time or anything like that)... but if we ever support multiple flux columns in the data entry, that may give a good chance to have dropdowns for which specific data columns are used as input.

lcviz/plugins/frequency_analysis/frequency_analysis.py
kecnry: Would you want to provide this is a single float and/or list?  We don't have a great way to support custom lists as inputs in the UI, except to parse the string and split on commas/spaces, which is why I skipped adding this in the first pass so far.

lcviz/plugins/frequency_analysis/frequency_analysis.py
kecnry: `auto_range` does not pass anything on to `lightkurve` and therefore allows astropy to compute defaults under-the-hood (`minimum` and `maximum` are ignored and not shown in the UI).

The defaults for `minimum` and `maximum` when `auto_range` is disabled are currently hardcode at 1-10 days, but it would be fairly easy to have that adjust to the input data (the complication then of course is that if the user intentionally sets the value of one of these but then changes the input data, they may not have intended that to override the value they intentionally set).

> Also, what are the units on these axes?

Fair, I'll add units into the axes labels (but the intention is that the units for period/frequency obey display units once that is ported over to the scatter viewer and lcviz).

lcviz/plugins/frequency_analysis/frequency_analysis.py
kecnry: Good catch!  This is likely a caching but of the underlying input dataset - I'll try to track it down and fix it.

lcviz/plugins/frequency_analysis/frequency_analysis.py
bmorris3: One float at a time is fine – it's an interactive workflow, and if you need, you can change it and look at the update 👍🏻

lcviz/plugins/frequency_analysis/frequency_analysis.py
bmorris3: I'm not sure what's best either. I will always wish we could use introspection to auto-populate traitlets and the UI from a call signature. 🤔

lcviz/plugins/frequency_analysis/frequency_analysis.py
kecnry: (units added to x-axis, I think the rest might make sense to defer until we have a better idea of how exactly we want the defaults to act)

lcviz/plugins/frequency_analysis/frequency_analysis.py
kecnry: this is actually caused by https://github.com/bqplot/bqplot/issues/1614 (we clear the periodogram briefly as the light curve data is being replaced which causes the figure to "crash").  The workaround in the meantime will likely need to be in jdaviz, not here.

PR #33, opened 2023-07-31
update is_active implementation from jdaviz 3.6 release

kecnry:
this updates lcviz to make use of https://github.com/spacetelescope/jdaviz/pull/2295


bmorris3 (approved):
Let's set `is_active=True` for the `is_active` PR.

PR #34, opened 2023-08-01
Load multiple data with different start times

bmorris3:
### Description

This PR adds support for loading multiple data entries into one LCviz instance. Changes required included:
* properly managing the "reference time" of the first loaded data
* consistent linking between temporal axes across time/phase viewers
* using consistent glue `ComponentID`s for time components in the LightCurve translator and in the Ephemeris plugin
* overriding the `viewer.state.reset_limits` method for light curves
* refactoring the `TimeCoordinate` class, which acts as the WCS-like coordinate object for light curves

I also updated the example notebook with Ephemeris and Flatten use cases with two quarters of the HAT-P-11 Kepler light curve.

glue-jupyter's bqplot interface has a `density_map` feature which renders a 2D histogram of datapoints when there are many datapoints in a viewer. In this PR, I turn that feature off for LCviz, though it may be useful to make this toggle easier to access in the future.

### DIFF
--- a/lcviz/parsers.py
+++ b/lcviz/parsers.py
@@ -43,17 +43,7 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
     if flux_origin is not None:
         new_data_label += f'[{flux_origin}]'
 
-    # grab the first-found reference time in the data collection:
-    ff_reference_time = None
-    for existing_data in app.data_collection:
-        if hasattr(existing_data, 'meta') and 'reference_time' in existing_data.meta:
-            ff_reference_time = existing_data.meta.get('reference_time', None)
-            if ff_reference_time is not None:
-                break
-
-    # convert to glue Data manually, so we may edit the `dt` component if necessary:
-    handler, _ = data_translator.get_handler_for(light_curve)
-    data = handler.to_data(light_curve, reference_time=ff_reference_time)
+    data = _data_with_reftime(app, light_curve)
     app.add_data(data, new_data_label)
 
     if show_in_viewer:
@@ -64,3 +54,17 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
         if ephem_plugin is not None:
             for viewer_id in ephem_plugin._obj.phase_viewer_ids:
                 app.add_data_to_viewer(viewer_id, new_data_label)
+
+
+def _data_with_reftime(app, light_curve):
+    # grab the first-found reference time in the data collection:
+    ff_reference_time = None
+    for existing_data in app.data_collection:
+        if hasattr(existing_data, 'meta') and 'reference_time' in existing_data.meta:
+            ff_reference_time = existing_data.meta.get('reference_time', None)
+            if ff_reference_time is not None:
+                break
+
+    # convert to glue Data manually, so we may edit the `dt` component if necessary:
+    handler, _ = data_translator.get_handler_for(light_curve)
+    return handler.to_data(light_curve, reference_time=ff_reference_time)
--- a/lcviz/parsers.py
+++ b/lcviz/parsers.py
@@ -1,6 +1,6 @@
 import os
 from astropy.io import fits
-from glue.core.link_helpers import LinkSame
+from glue.config import data_translator
 from jdaviz.core.registries import data_parser_registry
 from lightkurve import LightCurve, KeplerLightCurve, TessLightCurve
 from lightkurve.io.detect import detect_filetype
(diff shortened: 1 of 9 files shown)


lcviz/parsers.py
```diff
@@ -59,3 +54,17 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
         if ephem_plugin is not None:
             for viewer_id in ephem_plugin._obj.phase_viewer_ids:
                 app.add_data_to_viewer(viewer_id, new_data_label)
+
+
+def _data_with_reftime(app, light_curve):
+    # grab the first-found reference time in the data collection:
```
kecnry, line 60: what are the implications of this on the order that data is loaded into the app?

lcviz/parsers.py
```diff
@@ -59,3 +54,17 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
...
+    ff_reference_time = None
+    for existing_data in app.data_collection:
+        if hasattr(existing_data, 'meta') and 'reference_time' in existing_data.meta:
+            ff_reference_time = existing_data.meta.get('reference_time', None)
+            if ff_reference_time is not None:
+                break
+
+    # convert to glue Data manually, so we may edit the `dt` component if necessary:
+    handler, _ = data_translator.get_handler_for(light_curve)
+    return handler.to_data(light_curve, reference_time=ff_reference_time)
```
kecnry, line 70: would it be possible to get through the for-loop with `ff_reference_time` still `None`?  Does passing that on to the handler cause any issues?

lcviz/utils.py
```diff
@@ -27,84 +29,72 @@ def __init__(self, times, reference_time=None, unit=u.d):
...
-        Returns
-        -------
-        """
         if len(world) > 1:
             raise ValueError('TimeCoordinates is a 1-d coordinate class '
                              'and only accepts a single scalar or array to convert')
-        return np.interp(world[0], self._values.value, self._index,
-                         left=np.nan, right=np.nan)
+        return interp1d(
+            self._values.value, self._index, fill_value='extrapolate'
```
kecnry, line 51: what cases would require extrapolation?  Does that make any assumption based on cadence, etc?

lcviz/viewers.py
```diff
@@ -166,11 +171,19 @@ def add_data(self, data, color=None, alpha=None, **layer_state):
         """
         result = super().add_data(data, color, alpha, **layer_state)
 
+        for layer in self.layers:
+            # optionally render as a density map
+            layer.state.density_map = self.density_map
```
kecnry, line 176: if we don't currently have plans to use `density_map`, maybe we should just hardcode to False for now?  Or does passing through the kwarg actually have a use-case already?

notebooks/LCvizExample.ipynb
kecnry, line 1: I think we want to avoid the magic in the for-loop for the sake of the example notebook (and we might even change that behavior and allow flattened-results as input to the flattening plugin, which would then break this notebook as-is).

kecnry (commented):
This seems to address a whole bunch of different small problems - thanks!  Just a few questions on implications of everything, but the code changes seem reasonable to me.

lcviz/parsers.py
bmorris3: The main implication is that the origin of the time axis is set by the first Data in the data collection. If there's >1 Data, whatever's first in the collection sets the origin. 

It's possible this could lead to some confusion in the future, e.g., from data deletion. If you load three quarters, and delete the first loaded quarter, the next two quarters got their reference times from the first-loaded data, so the time origin doesn't change. 

An alternative/enhanced implementation to all of this is to support global reference time changes, so that you can set the time origin to any time you want. We could put it in an attr on the helper and propagate its updates into the Data components.

bmorris3: I like that proposition more and more as I think about it. We could allow the user to give a time to the Plot Options plugin.

lcviz/parsers.py
bmorris3: If `ff_reference_time is None`, the `TimeCoordinate` class uses the first time in the LC as the time origin: https://github.com/spacetelescope/lcviz/pull/34/files/f3df2d3e203df20f65be1013317b8bed3d81e91b#diff-0dd872fcfa6b17e68d28e790b9f1859ba353e985725131fc34fa72b7c90258f5R36

lcviz/utils.py
bmorris3: Extrapolation is needed to support continuous coord-info for non-overlapping light curves. If the world coordinate becomes `nan` before the first cadence, glue can't plot multiple light curves correctly. The default extrapolation is linear from the first/last two times in the LC for the left/right sides.

lcviz/viewers.py
bmorris3: I liked using it in the flux-vs-phase viewer with ephemeris, and it's really the only scalable way to do short cadence light curves. I think we'll need to figure out how/when to use `density_map` before long.

notebooks/LCvizExample.ipynb
bmorris3: Good idea. Updated the notebook here: ae5ebec

lcviz/parsers.py
kecnry: Ok, I think we actually already had that in mind [:cat:](https://jira.stsci.edu/browse/JDAT-3369), so let's defer for now.

kecnry (approved):

PR #38, opened 2023-08-22
update to use skip_if_no_updates_since_last_active

kecnry:
this accompanies https://github.com/spacetelescope/jdaviz/pull/2386 and updates the flatten and binning plugins to make use of this new decorator, which should avoid any flickering for expensive cases.

This will require jdaviz 3.7 ~(once released - marking as draft until then)~.  ~The binning plugin from #28 will also require updates (either as part of that PR or this rebased on top of that once merged)~

### DIFF
--- a/lcviz/plugins/flatten/flatten.py
+++ b/lcviz/plugins/flatten/flatten.py
@@ -6,7 +6,8 @@ from jdaviz.core.custom_traitlets import FloatHandleEmpty, IntHandleEmpty
 from jdaviz.core.events import ViewerAddedMessage
 from jdaviz.core.registries import tray_registry
 from jdaviz.core.template_mixin import (PluginTemplateMixin,
-                                        DatasetSelectMixin, AddResultsMixin)
+                                        DatasetSelectMixin, AddResultsMixin,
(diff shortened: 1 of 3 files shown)


lcviz/plugins/flatten/flatten.py
```diff
@@ -159,16 +160,25 @@ def _clear_marks(self):
...
+            mark.visible = visible
+
+        if visible and event.get('name') in ('is_active', 'show_live_preview'):
+            # then the marks themselves need to be updated
+            self._live_update(event)
+
+    @observe('dataset_selected',
              'window_length', 'polyorder', 'break_tolerance',
              'niters', 'sigma')
+    @skip_if_no_updates_since_last_active()
```
bmorris3, line 180: Unimportant: if a decorator takes no arguments, does it need to be followed by `()`?

bmorris3 (approved):
Looks good!

lcviz/plugins/flatten/flatten.py
kecnry: This decorator does accept arguments, its just using the default in this case (so here it is needed).

PR #35, opened 2023-08-03
Moving viewer reference renaming upstream to jdaviz

bmorris3:
https://github.com/spacetelescope/jdaviz/pull/2338 implements viewer reference name updates in jdaviz, removing the need for that functionality in LCviz. 

This PR removes the duplicate code, and calls the jdaviz versions. This PR should not be merged until https://github.com/spacetelescope/jdaviz/pull/2338 is on main.

[🐱](https://jira.stsci.edu/browse/JDAT-3680)

PR #39, opened 2023-08-23
increase test coverage

kecnry:
This PR increases overall test coverage from 85.65% to 92.95% (and simplifies some logic in the light curve parser).

NOTE: uncovered lines in `events.py` and `template_mixin.py` are scheduled to be moved upstream by #35 and #36, respectively, which should then increase percent coverage even further.

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -31,7 +31,7 @@ def _get_range_subset_bounds(self, subset_state, *args, **kwargs):
         # TODO: use display units once implemented in Glue for ScatterViewer
         # units = u.Unit(viewer.state.x_display_unit)
         units = u.Unit(viewer.time_unit)
-    else:
+    else:  # pragma: no cover
         raise ValueError("Unable to find time axis units")
 
     region = reference_time + u.Quantity([subset_state.lo * units, subset_state.hi * units])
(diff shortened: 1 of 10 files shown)


lcviz/parsers.py
```diff
@@ -17,21 +15,11 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
...
-
-        # get the constructor for this type of light curve:
-        filetype_to_cls = {
-            'KeplerLightCurve': KeplerLightCurve,
-            'TessLightCurve': TessLightCurve
-        }
-        cls = filetype_to_cls[filetype]
         # read the light curve:
-        light_curve = cls.read(file_obj)
+        light_curve = lightkurve.read(file_obj)
```
kecnry, line 19: @bmorris3 - any reason why we can't rely on lightkurve functionality here?

lcviz/helper.py
```diff
@@ -31,7 +31,7 @@ def _get_range_subset_bounds(self, subset_state, *args, **kwargs):
         # TODO: use display units once implemented in Glue for ScatterViewer
         # units = u.Unit(viewer.state.x_display_unit)
         units = u.Unit(viewer.time_unit)
-    else:
+    else:  # pragma: no cover
```
rosteen, line 34: Well, that's one way to improve coverage 😉

rosteen (approved):
LGTM, tests pass, nothing objectionable in the code. Nice to simplify the parser.

PR #41, opened 2023-08-30
Flatten: separate previews for flattened and trend

kecnry:
**NOTE**: This PR should be rebased after #38 is merged (currently builds on top of it to avoid needing to re-implement), and therefore needs to wait for jdaviz 3.7 to be released.

This PR separates the plugin-settings switch for live-previews in the flattening plugin to control the (unnormalized) flattened light curve and the trend separately.  In cases where there is not much flattening to be done, it is difficult to see the trend and ensure it is correct if the flattened light curve is directly on top.


lcviz/plugins/flatten/flatten.py
```diff
@@ -43,6 +46,7 @@ class Flatten(PluginTemplateMixin, DatasetSelectMixin, AddResultsMixin):
     uses_active_status = Bool(True).tag(sync=True)
 
     show_live_preview = Bool(True).tag(sync=True)
+    show_trend_preview = Bool(True).tag(sync=True)
```
bmorris3, line 49:
```suggestion
    show_trend_preview = Bool(False).tag(sync=True)
```
Take or leave this suggestion, but I think the default should be less cluttered and redundant. Especially since the trend color and live preview color are identical, there's a lot to see in the viewer if this is enabled by default.

lcviz/plugins/flatten/flatten.py
```diff
@@ -26,7 +26,10 @@ class Flatten(PluginTemplateMixin, DatasetSelectMixin, AddResultsMixin):
     Only the following attributes and methods are available through the
     public plugin API.
 
-    * ``show_live_preview``
+    * ``show_live_preview`` : bool
+        Whether to show the live-preview of the (unnormalized) flattened light curve
+    * ``show_trend_preview`` : bool
+        Whether to show the live-preview of the removed trend
```
bmorris3, line 32:
```suggestion
        Whether to show the live-preview of the trend curve used to flatten the light curve
```

bmorris3 (approved):
Looks good!

lcviz/plugins/flatten/flatten.py
kecnry: usually I actually prefer to see the trend over the preview 🤷‍♂️   Let's leave for now and revisit as we use it more.

PR #45, opened 2023-09-05
enable subset freezing

kecnry:
This enables support from ~https://github.com/spacetelescope/jdaviz/pull/2421~ https://github.com/spacetelescope/jdaviz/pull/2462 into lcviz's subset plugin.

https://github.com/spacetelescope/lcviz/assets/877591/83ca9d40-5a13-40ff-9675-5629c1db98e3

~This would require that PR to be merged (along with the necessary fixes for get_subsets and the subset plugin), but should not require it to be released in order to move forward with this PR.~

~This can be merged without jdaviz 3.7 released, but will require testing with jdaviz after that PR is merged (lcviz should work with _and_ without, but this feature will only be shown if you have jdaviz _with_ the upstream PR).~


bmorris3 (approved):
Neat! This is handy. For example, you can wrap at 0.5, selecting a subset bracketing zero, freeze the subset, and re-wrap at 1:

Screen Shot 2023-09-22 at 13 42 22

PR #46, opened 2023-09-05
DOC: update documentation

kecnry:
This PR:
* updates the docs theme to use the same as jdaviz
* adds information on the index page briefly explaining lcviz and linking to jdaviz and lightkurve
* copies (relevant) installation information from jdaviz
* exposes API docs for plugins
* links from each plugin to jdaviz/lightkurve, when applicable
* provides user-API snippets for each of the plugins

This does not intend to be a comprehensive/final update to the docs, but just to get in shape for an initial release.  Areas that could still use improvement include:
* add lcviz logo (once created)
* screenshots/recording on splash page
* extended descriptions for each plugin
* development docs (anything that is lcviz-specific, otherwise point to jdaviz docs)

[rendered docs](https://lcviz--46.org.readthedocs.build/en/46/)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,128 @@
+
+.. _install:
+
+Install
+=======
+
+.. note::
+
+    ``lcviz`` is undergoing constant development. We encourage users to always update
+    to the latest version. In general, it is good practice to install the development
+    version following the instructions below as full released versions may lag behind.
+
+User Installation
+-----------------
+
+Create Your Local Environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some of Jdaviz's dependencies require non-Python packages to work
+(particularly the front-end stack that is part of the Jupyter ecosystem).
+We recommend using `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_
+to easily manage a compatible Python environment for ``jdaviz``; it should work
+with most modern shells, except CSH/TCSH.
+
+You may want to consider installing ``lcaviz`` in a new virtual or conda environment
+to avoid version conflicts with other packages you may have installed, for example:
+
+.. code-block:: bash
+
+    conda create -n lcviz-env python=3.9
+    conda activate lcviz-env
+
+Pip Install
+^^^^^^^^^^^
+
+As noted above, we typically recommend installing the latest development version:
+
+.. code-block:: bash
+
+    pip install git+https://github.com/spacetelescope/lcviz --upgrade
+
+A normal install will also work by installing the latest release version:
+
+.. code-block:: bash
+
+    pip install lcviz --upgrade
+
+Common Issues
+^^^^^^^^^^^^^
+
+If you encounter problems while following these installation instructions,
+please consult :ref:`known installation issues <known_issues_installation>`.
+
+Note that ``lcviz`` requires Python 3.9 or newer. If your ``pip`` corresponds to an older version of
+Python, it will raise an error that it cannot find a valid package.
+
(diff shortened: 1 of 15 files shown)


docs/installation.rst
```diff
@@ -0,0 +1,128 @@
...
+
+    pip install watchdog
+
+After installing ``watchdog``, to use it, add the following to the top
+of a notebook:
+
+.. code-block:: python
+
+    from lcviz import enable_hot_reloading
+    enable_hot_reloading()
```
bmorris3, line 128: More dev docs – can we reiterate this in the dev docs?

https://github.com/spacetelescope/jdaviz/blob/fefee34a38ce2136e578ca2d806d69e2c9ff4296/setup.py#L48-L50

I'm still in the habit of running `tox -e build_docs`, which won't work here (yet).

docs/plugins.rst
```diff
@@ -9,9 +9,29 @@ Metadata Viewer
...
+      lc = search_lightcurve("HAT-P-11", mission="Kepler",
+                             cadence="long", quarter=10).download().flatten()
+      lcviz = LCviz()
+      lcviz.load_data(lc)
+      lcviz.show()
+
+      metadata = lcviz.plugins['Metadata']
+      print(f"dataset choices: {metadata.dataset.choices}")
+      metadata.dataset = metadata.dataset.choices[0]
+      print(metadata.metadata)
```
bmorris3, line 28: Inline comments would be useful here to explain what we're printing and why.

docs/plugins.rst
```diff
@@ -21,13 +41,44 @@ Plot Options
...
+                             cadence="long", quarter=10).download().flatten()
+      lcviz = LCviz()
+      lcviz.load_data(lc)
+      lcviz.show()
+
+      po = lcviz.plugins['Plot Options']
+      print(f"viewer choices: {po.viewer.choices}")
+      po.viewer = po.viewer.choices[0]
+      print(f"layer choices: {po.layer.choices}")
+      po.layer = po.layer.choices[0]
```
bmorris3, line 60: Inline comments would be useful here to explain what we're printing and why.

docs/plugins.rst
```diff
@@ -52,9 +129,30 @@ This plugin allows for flattening the light curve by removing trends.  By defaul
...
+                             cadence="long", quarter=10).download()
+      lcviz = LCviz()
+      lcviz.load_data(lc)
+      lcviz.show()
+
+      flatten = lcviz.plugins['Flatten']
+      flatten.open_in_tray()
+      flatten.polyorder = 4
+      flattened_lc = flatten.flatten(add_data=True)
+      print(flattened_lc)
```
bmorris3, line 148: Inline comments would be useful here to explain what we're printing and why.

bmorris3 (commented):
There are several tests called `test_docs_snippets` in this PR that don't seem to check anything – should those be in this PR? Seems like we're reengineering doctests?

docs/installation.rst
kecnry: I'll add dev docs to the list of follow-up as well.  I don't want to just duplicate jdaviz docs in that case... but if there's anything lcviz-specific, we should definitely include it.

docs/plugins.rst
kecnry: these are intended to be somewhat "self-instructing" snippets... I don't want to get into the territory of narrative notebooks.  Do you have specific suggestions for what should be added?

kecnry:
Merging as-is and can always continue to improve the docs.  Thanks for the ideas!

PR #51, opened 2023-09-15
Binning: handle n_bins being empty or <= 0

kecnry:
This PR handles the case where the user empties the input for number of bins or sets it to an invalid value.  The live-preview now clears without a traceback and the "bin" button is disabled.  The input validation shows red text describing the reason the input is invalid (unchanged by this PR).


bmorris3 (approved):
This works great, approving.

One thing to consider adding to the validation (now or in the future) is to limit `n_bins <= len(y)`. Doesn't make much sense otherwise.

PR #52, opened 2023-09-22
update installation of astropy dev wheel

kecnry:
See https://github.com/spacetelescope/jdaviz/pull/2358 (thanks @pllim!)


pllim:
How much do you care about testing with numpy dev? Because your log shows that matplotlib is preventing numpy dev from installing. You might also need:

* #2472

kecnry:
not too worried about that right now, we don't use much numpy internally, so jdaviz numpy-dev tests should catch anything... but definitely something we should consider once we're getting closer to a release.  Thanks!

PR #54, opened 2023-09-22
Alternate upstream renaming implementation

bmorris3:
Compare with #35.


kecnry:
jdaviz 3.7.1 is released, so we can move forward with #35 instead.

PR #55, opened 2023-10-19
Use upstream updates to plugin plots

kecnry:
This updates the plot in frequency analysis to account for changes upstream in https://github.com/spacetelescope/jdaviz/pull/2498 (which adds tool and pan/zoom support), and in doing so requires jdaviz 3.8.

Note that this also means that without this PR, the former implementation _may_ be broken once jdaviz 3.8 is released.

~**Waiting for:** jdaviz 3.8~

### DIFF
--- a/lcviz/plugins/frequency_analysis/frequency_analysis.py
+++ b/lcviz/plugins/frequency_analysis/frequency_analysis.py
@@ -178,7 +178,9 @@ class FrequencyAnalysis(PluginTemplateMixin, DatasetSelectMixin, PlotMixin):
 
         per = self.periodogram
         if per is not None:
-            self.plot._update_data('periodogram', x=getattr(per, self.xunit_selected), y=per.power.value)
+            self.plot._update_data('periodogram',
+                                   x=getattr(per, self.xunit_selected),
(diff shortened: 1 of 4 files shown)


lcviz/plugins/frequency_analysis/frequency_analysis.py
```diff
@@ -69,8 +70,11 @@ def __init__(self, *args, **kwargs):
                                            selected='xunit_selected',
                                            manual_options=['frequency', 'period'])
 
-        self.plot.add_line('line', color='gray', marker_size=12)
         self.plot.figure.axes[1].label = 'power'
+        self.plot.figure.fig_margin = {'top': 60, 'bottom': 60, 'left': 65, 'right': 15}
```
bmorris3, line 74: Should `left` be larger?

Screen Shot 2023-11-30 at 12 13 26

bmorris3: Correction: the line above is not responsible for the figure/viewer axis label spacing.

bmorris3 (approved):

PR #56, opened 2023-11-09
flatten/binning: handle slow live-updates for large data

kecnry:
This PR introduces a check on the time that the live-updates take, and if exceeding a certain threshold, temporarily disables the live-preview updates and shows an alert prompting to either manually update the live-preview or disable the live-preview marks.

This also makes use of a [change upstream in jdaviz](https://github.com/spacetelescope/jdaviz/pull/2560) to show a spinner in the "apply" buttons when the plugin is computing which temporarily disables clicking the button.  For cases where binning/flattening is slow to run, this gives immediate feedback that the button was clicked and doesn't need to be clicked again.  This doesn't need to be blocked by pinning a new release of jdaviz, ~but should make sure that PR gets merged first~.

The code here can be simplified slightly by using the `@with_spinner()` decorator from https://github.com/spacetelescope/jdaviz/pull/2560, but that can be a follow-up once that is in a jdaviz release.

Note: this video was before some style changes in https://github.com/spacetelescope/jdaviz/pull/2560 - the check marks are now gone and will show a spinner, if running on jdaviz main.  If not running on main, the "bin" button will not react, but everything else will still work.

https://github.com/spacetelescope/lcviz/assets/877591/ac98e261-d4d6-43f7-990a-b8db96d1537d

### DIFF
--- a/lcviz/plugins/binning/binning.py
+++ b/lcviz/plugins/binning/binning.py
@@ -1,5 +1,7 @@
+import numpy as np
+from time import time
 from astropy.time import Time
-from traitlets import Bool, observe
+from traitlets import Bool, Float, observe
 from glue.config import data_translator
 
 from jdaviz.core.custom_traitlets import IntHandleEmpty
@@ -45,6 +47,10 @@ class Binning(PluginTemplateMixin, DatasetSelectMixin, EphemerisSelectMixin, Add
 
     n_bins = IntHandleEmpty(100).tag(sync=True)
     bin_enabled = Bool(True).tag(sync=True)
+
+    last_live_time = Float(0).tag(sync=True)
+    previews_temp_disable = Bool(False).tag(sync=True)
+    spinner = Bool(False).tag(sync=True)
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


lcviz/plugins/binning/binning.py
```diff
@@ -189,6 +200,10 @@ def _live_update(self, event={}):
                 mark.times = []
                 mark.update_xy(times, lc.flux.value)
 
+        self.last_live_time = np.round(time() - start, 2)
+        if self.last_live_time > 0.3:
```
bmorris3, line 204: Can we make the threshold a traitlet, and expose it in the user API? It'd be great if a user could assert: "I'm sick of that warning, I'll gladly wait, leave me alone." The user could set the number to `10_000`, get coffee, and come back without a warning.

bmorris3 (approved):
Looks good, works for me. One small revision requested, but not required.

lcviz/plugins/binning/binning.py
kecnry: I like this idea, but even though we aren't technically committing to API yet, we might soon, and I think it might make sense to play with it and converge more on the feature (if necessary) before we make it customizable.

bmorris3: I'm happy with that. 👍🏻

PR #62, opened 2023-11-14
Add folding period to ephemeris viewers axis label

bmorris3:
It's easy to forget what period you have set in the Ephemeris plugin, and it's a bit annoying to need to open the plugin if it isn't already visible. This confusion is multiplied when you have multiple ephemerides (with distinct periods).

This PR adds the period to the x-axis label in phase viewers. Here's an example for RR Lyra:

Screen Shot 2023-11-13 at 20 53 35

The x axis is probably the best place to put the period label for now. I tried using bqplot's `title` attr, but it only sits at the top-center. See the right-most phase viewer here: 

Screen Shot 2023-11-13 at 21 06 48

Often the data will sit beneath that label, so I prefer the x label.

### DIFF
--- a/lcviz/viewers.py
+++ b/lcviz/viewers.py
@@ -219,8 +219,13 @@ class PhaseScatterView(TimeScatterView):
     def _set_plot_x_axes(self, dc, component_labels, light_curve):
         # setting of y_att will be handled by ephemeris plugin
         self.state.x_att = dc[0].components[component_labels.index(f'phase:{self.ephemeris_component}')]  # noqa
-        self.figure.axes[0].label = 'phase'
         self.figure.axes[0].num_ticks = 5
+
+        ephem = self.jdaviz_helper.plugins.get('Ephemeris', None)
+        if ephem:
+            self.figure.axes[0].label = f'phase (P = {ephem.period:.2g} d)'
+        else:
+            self.figure.axes[0].label = 'phase'
 
     def times_to_phases(self, times):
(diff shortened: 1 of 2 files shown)


lcviz/viewers.py
```diff
@@ -219,9 +219,14 @@ def ephemeris_component(self):
     def _set_plot_x_axes(self, dc, component_labels, light_curve):
         # setting of y_att will be handled by ephemeris plugin
         self.state.x_att = dc[0].components[component_labels.index(f'phase:{self.ephemeris_component}')]  # noqa
-        self.figure.axes[0].label = 'phase'
         self.figure.axes[0].num_ticks = 5
 
+        ephem = self.jdaviz_helper.plugins.get('Ephemeris', None)
+        if ephem:
+            self.figure.axes[0].label = f'phase (P = {ephem.period:.2g} d)'
```
kecnry, line 226: I wonder if this rounding could be deceptive.... can we use tex to show `~=`?

We could also add a settings expansion menu and make this a togglable option (having `phase:ephemeris_component` might also be an option people want in some cases, and actually briefly happens sometimes since that is the name of the data component).  It would be _really_ cool if we had some presets for the labels but also allowed custom formatting 😉

kecnry: Getting back to this - I think for now this is a good improvement, _if_ we can use tex to avoid rounding confusion.  Let's defer anything else to potential follow-up work [:cat:](https://jira.stsci.edu/browse/JDAT-3992).

PR #64, opened 2023-11-16
expose adopt_period_at_max_power through ephem API

kecnry:
as it says in the title, useful for notebook workflows, but just didn't want the diff buried in a notebook PR.


bmorris3 (approved):
10/10 yes.

PR #63, opened 2023-11-16
RR Lyrae example notebook

bmorris3:
While we discuss offline about supporting too many demo notebooks like #59, I'm also opening this PR with the RR Lyrae example, just so it's logged somewhere. We can close this and #59 if we want to suggest these NBs elsewhere (jdat notebooks?) in the future.

~**Waiting for**: jdaviz 3.8~

PR #65, opened 2023-11-16
demo notebook for triple eclipsing system

kecnry:
This is a demo notebook illustrating using the flatten and ephemeris plugins for KIC 2835289 - a triple eclipsing system with an ellipsoidal variable as the inner-binary.  It demonstrates API commands to manage multiple ephemerides and could be extended with subset workflows or coloring by phase (avoiding too much overlap with #63).

This requires:
* #64 
* https://github.com/spacetelescope/jdaviz/pull/2563 

~**Waiting for**: jdaviz 3.8 (for `close_in_tray` and API access to set viewer limits)~

### DIFF
--- a/notebooks/LCvizDemoMultEphem.ipynb
+++ b/notebooks/LCvizDemoMultEphem.ipynb
@@ -109,7 +109,7 @@
    "source": [
     "## Phase-Folding\n",
     "\n",
-    "Similarly, we'll use the \"Ephemeris\" plugin to create two ephemerides: one for the inner and one for the outer orbit.  As these ephemerides are created, new phase viewers will dynamically be created to show the same light curve in the phase-space as defined by that ephemeris.  Similarly to the flattening plugin, this call all also be done interactively through the \"Ephemeris\" plugin, with the phase-folding updated in real-time."
+    "We can use the \"Ephemeris\" plugin to visualize two periodic features in this light curve: one period for the inner binary, and another period for the outer star's orbit.  As each ephemeris is added, a new phase viewer will be created to show the light curve folded on that period.  As with the Flatten plugin above, the UI allows for live-updating interactive workflows through the \"Ephemeris\" plugin."
    ]
   },
   {


notebooks/LCvizDemoMultEphem.ipynb
```diff
@@ -0,0 +1,269 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f90285a3-ef1d-4491-94f8-05c2f523c7d6",
+   "metadata": {},
+   "source": [
+    "## Flattening / Detrending\n",
+    "\n",
+    "Here we'll use the plugin API to both open the plugin UI, set the inputs, and ultimately flatten the light curve.  To reproduce this interactively, open the \"Flatten\" plugin in the plugin tray (click the three horizontal bars in the top-right of lcviz), and set the corresponding inputs in the plugin UI.  Note that this provides immediate feedback as to how those inputs affect the trend as well as the final flattened light curve."
```
bmorris3, line 37:
```suggestion
    "Here we'll use the API to open the plugin UI, set the inputs, and flatten the light curve.  To reproduce this workflow interactively, open the \"Flatten\" plugin in the plugin tray (click the three horizontal bars in the top-right of lcviz), and set the corresponding inputs in the plugin UI. Modifying the parameters in the UI will produce live-previews of the detrended light curve."
```

notebooks/LCvizDemoMultEphem.ipynb
```diff
@@ -0,0 +1,269 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fcbdd24a-52e6-467d-a3ce-570af1fb0bf6",
+   "metadata": {},
+   "source": [
+    "## Phase-Folding\n",
+    "\n",
+    "Similarly, we'll use the \"Ephemeris\" plugin to create two ephemerides: one for the inner and one for the outer orbit.  As these ephemerides are created, new phase viewers will dynamically be created to show the same light curve in the phase-space as defined by that ephemeris.  Similarly to the flattening plugin, this call all also be done interactively through the \"Ephemeris\" plugin, with the phase-folding updated in real-time."
```
bmorris3, line 112:
```suggestion
    "We can use the \"Ephemeris\" plugin to visualize two periodic features in this light curve: one period for the inner binary, and another period for the outer star's orbit.  As each ephemeris is added, a new phase viewer will be created to show the light curve folded on that period.  As with the Flatten plugin above, the UI allows for live-updating interactive workflows through the \"Ephemeris\" plugin."
```

notebooks/LCvizDemoMultEphem.ipynb
```diff
@@ -0,0 +1,269 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "8f932e3b-4c43-4295-bd9f-f0cbce75fbf0",
+   "metadata": {},
+   "source": [
+    "# Phase-Folding on Multiple Ephemerides\n",
+    "\n",
+    "This notebook loads a triple eclipsing system (KIC 2835289) containing an inner-ellipsoidal variable, flattens the light curve, and phase-folds on both orbital periods."
```
bmorris3, line 10: I took a look at your paper and at this light curve, and I'm still confused by the geometry. Do all three objects have eclipses in the light curve? If not, calling this a "triple eclipsing system" is misleading. You could say "eclipsing triple system." 

But looking at the light curve, the ellipsoidal variations are visible during the eclipse, which would happen if the ellipsoidal binary system is in orbit around a much brighter primary. Am I crazy?

Here's an alternative consistent with my (possibly crazy) understanding:

```suggestion
    "This notebook loads the Kepler light curve for triple system (KIC 2835289) containing a binary with ellipsoidal variations which eclipses a brighter star. We use lcviz to flatten the light curve, and phase-fold on the ellipsoidal period and the binary orbital period around the primary."
```

kecnry: eclipsing triple I suppose is a little more accurate (although often ellipsoidal variables are lumped in with eclipsing binaries even though they don't technically eclipse), but I am nearly positive it is the third star that eclipses the binary and not vice versa.

PR #70, opened 2023-12-07
Release workflow

kecnry:
This PR enables `lcviz.test` and adapts the release workflow used in jdaviz.

- [x] ~need to setup the secret token in the repo before first release!~ (switched to hopefully working with trusted publisher without the need for a token)
- [ ] do we want changelog enforcement going forward?  (That can always be added later though)


bmorris3 (approved):
Hooray! Let's do it!

PR #74, opened 2023-12-13
clone viewer tool

kecnry:
This PR implements a toolbar tool capable of cloning the current viewer (keeping axes limits, data loaded/visible, and plot options - except for bug noted below) in a way that still works seemlessly with the ephemeris plugin.

https://github.com/spacetelescope/lcviz/assets/877591/30cc9f8c-6fd8-48c0-921c-c0e2cbc7272f

https://github.com/spacetelescope/lcviz/assets/877591/12eb05a9-0c26-4c37-ace3-1360b505fb4d

NOTE: known bug with layer options reverting when adding to any new viewer (was not introduced here, but this definitely exposes it more readily)

**TODO**:
- [x] test coverage
- [ ] custom icon (could be deferred)
- [ ] consider whether this would be moved upstream
- [ ] consider exposing in viewer API (will be easier once/if moved upstream)


CHANGES.rst
```diff
@@ -1,5 +1,10 @@
-0.1.0 - unreleased
+0.2.0 - unreleased
 ------------------
 
+* Clone viewer tool. [#74]
+
+0.1.0
+-----
```
bmorris3, line 7:
```suggestion
0.1.0 (12-14-2023)
------------------
```

lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -308,7 +308,8 @@ def vue_period_double(self, *args):
         self.period *= 2
 
     def _check_if_phase_viewer_exists(self, *args):
-        self.phase_viewer_exists = self.phase_viewer_id in self.app.get_viewer_ids()
+        viewer_base_refs = [id.split('[')[0] for id in self.app.get_viewer_ids()]
```
bmorris3, line 311: I'm OK with assuming there aren't other square brackets in the viewer ID. Just leaving a comment here to hopefully remind us both that, going forward, we're assuming that.

bmorris3 (approved):
This works great. LGTM.

lcviz/plugins/ephemeris/ephemeris.py
kecnry: yep, this is very much assuming we always stick to these syntax rules... but if we ever want viewer renaming, we can always move these things into attributes rather than parsing the viewer name itself, I just don't want to overengineer it until then.

PR #75, opened 2023-12-19
Adding TPF translator and parser

bmorris3:
Try it out with:

```python
import astropy.units as u
from astropy.time import Time
from lightkurve import search_lightcurve, search_targetpixelfile

from lcviz import LCviz

# download a Kepler TPF:
tpf = search_targetpixelfile('KIC 6922244', author="Kepler", quarter=9, cadence="long").download()

# download a TESS TPF:
# tpf = search_targetpixelfile('TIC 307210830', mission="TESS", cadence="long").download()


# Load the light curve into LCviz:
lcviz = LCviz()
lcviz.load_data(tpf)
lcviz.show()
```

### DIFF
--- a/lcviz/parsers.py
+++ b/lcviz/parsers.py
@@ -22,6 +22,10 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
     elif isinstance(file_obj, lightkurve.LightCurve):
         light_curve = file_obj
 
+    # load a LightCurve object:
+    elif isinstance(file_obj, lightkurve.targetpixelfile.KeplerTargetPixelFile):
+        light_curve = file_obj
+
     # make a data label:
     if data_label is not None:
         new_data_label = f'{data_label}'
--- a/lcviz/parsers.py
+++ b/lcviz/parsers.py
@@ -36,7 +36,7 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
 
     # handle flux_origin default
     flux_origin = light_curve.meta.get('FLUX_ORIGIN', None)  # i.e. PDCSAP or SAP
-    if flux_origin == 'flux' or (flux_origin is None and 'flux' in light_curve.columns):
+    if flux_origin == 'flux' or (flux_origin is None and 'flux' in getattr(light_curve, 'columns', [])):  # noqa
         # then make a copy of this column so it won't be lost when changing with the flux_column
         # plugin
         light_curve['flux:orig'] = light_curve['flux']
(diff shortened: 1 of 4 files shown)


lcviz/parsers.py
```diff
@@ -22,6 +22,10 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
     elif isinstance(file_obj, lightkurve.LightCurve):
         light_curve = file_obj
 
+    # load a LightCurve object:
+    elif isinstance(file_obj, lightkurve.targetpixelfile.KeplerTargetPixelFile):
+        light_curve = file_obj
```
kecnry, line 27: Is there any reason we can't merge these lines?  

```suggestion
    elif isinstance(file_obj, (lightkurve.LightCurve, lightkurve.KeplerTargetPixelFile):
        light_curve = file_obj
```

How about checking against the parent `TargetPixelFile` class (here and in utils as well) - or is some of this logic specific to only support Kepler?

kecnry (commented):
Oh I'm a bit surprised to see this actually plot the light curve... where is that happening?  Do we want that?  Or do we want to force the logic through the (eventual) extraction plugin?

kecnry:
TODO for this PR:
- [ ] generalize to all TPF files or at least Kepler, K2, Tess
- [ ] test coverage

Follow-up efforts (out-of-scope here):
- [ ] filter/exclude TPF entries from lc viewer data menus (might need logic or refactoring upstream)
- [ ] plugin-managed viewer to visualize cube (with appropriate data filtering in menu)
- [ ] extraction plugin with live-preview and pipeline defaults
- [ ] slice plugin with indicator in all time viewers

lcviz/parsers.py
bmorris3: As far as I can tell, there is only one feature (quality bitmask conversion) that is specific to the mission within the TPF translator. I can put in a check for the mission to use the correct bitmask converter, and then this should (hopefully) work for all TPF classes.

kecnry:
It seems that the resulting cube is transposed with respect to the convention in cubeviz.  Should we handle this here or elsewhere?

To test, try the following with the `tpf-viewer` branch, built on top of this:

```
from lightkurve import search_targetpixelfile
from lcviz import LCviz

tpf = search_targetpixelfile("HAT-P-11", mission="Kepler", cadence="long", quarter=10).download()

lcviz = LCviz()
lcviz.load_data(tpf)
lcviz.show()

lcviz.get_data().shape
# (4447, 22, 11) which is (time, x, y) whereas cubeviz uses (x, y, spectral_axis)
```

kecnry (approved):
Test coverage should fill in as this actually becomes usable 🤞

PR #77, opened 2023-12-28
Flux Column plugin

kecnry:
This PR implements a plugin for selecting the flux origin _per-dataset_ (note that this means that _each_ dataset chooses which column is treated as flux, and then _every_ viewer and plugin will use that column as input).

https://github.com/spacetelescope/lcviz/assets/877591/a8dc2fb9-b49b-4a09-9d71-52d575b77275

This also results in some necessary changes to the flatten plugin - which no longer writes to a new data entry, but rather appends a new column in the input light curve and selects that as the flux-origin by default.  This does make some workflows like comparing before and after flattening a little more cumbersome, but should be more natural for the more common use-cases:

https://github.com/spacetelescope/lcviz/assets/877591/14945642-f521-45ec-a055-f580dbf0abc9

Rendered plugin docs: https://lcviz--77.org.readthedocs.build/en/77/plugins.html#flux-column

TODO:
- [x] outputs from flattening do not play well because: the flux column is unitless, tripping the unit comparison logic, and other columns remain unflattened.  Would we want an option in flattening to append as a new column to the existing dataset instead of overwriting?  Should flattening plugin apply across all flux columns or drop remaining columns?
- [x] test workflow with binning.  Binning does bin each column (and has a different time column), so creating a new data entry here does make sense and works as-is.
- [x] clear cache on ANY dataset dropdown when flux column is changed (test with binning live-preview)

Potential follow-up:
- [ ] update or hide FLUX_ORIGIN from metadata plugin
- [ ] what if the user wants to plot these columns side-by-side (either in the same or separate viewers)?  Maybe a "clone" data (similar to clone viewer) would be appropriate in that case?


docs/plugins.rst
```diff
@@ -35,6 +35,41 @@ This plugin allows viewing of any metadata associated with the selected data.
     :ref:`Jdaviz Metadata Viewer <jdaviz:imviz_metadata-viewer>`
         Jdaviz documentation on the Metadata Viewer plugin.
 
+.. _flux-origin:
+
+Flux Origin
```
bmorris3, line 40: I know what you mean by `origin` but the term used throughout lightkurve is `column`. Is there a reason to introduce another term?

bmorris3 (commented):
Sorry for this small but annoying-to-address comment/question.

docs/plugins.rst
kecnry: hmmm I actually originally had column, but then changed it because its exposed as "FLUX_COLUMN" in the metadata and the string representation of the `LightCurve` object 🤔 

I guess [this](https://docs.lightkurve.org/reference/api/lightkurve.LightCurve.select_flux.html) and [this](https://docs.lightkurve.org/reference/api/lightkurve.SearchResult.download.html) both use "column", but [this tutorial](https://docs.lightkurve.org/tutorials/1-getting-started/using-light-curve-file-products.html?highlight=flux_origin) shows how many times origin occurs.

Ultimately it's not too difficult to change (now or before this makes a release - significantly more difficult after a release)... so we can either decide which one we think is better or exposed more publicly or ask if they'd ever consider moving to be more self-consistent within lightkurve and go with that?

bmorris3: Sounds good. Personally, I'd reserve "origin" for the origin of the axis, which might be, e.g., zero or one, depending on units and normalization.

lcviz/plugins/flux_column/flux_column.vue
```diff
@@ -0,0 +1,31 @@
...
+    />
+
+    <v-row>
+      <v-select
+        :menu-props="{ left: true }"
+        attach
+        :items="flux_column_items.map(i => i.label)"
+        v-model="flux_column_selected"
+        label="Flux Column"
+        hint="Select the column to adopt as flux."
```
bmorris3, line 25: I would love to put a link here to someone's docs that define the columns, so I looked around online to see if the Kepler/TESS/NASA docs have a good description of the column name definitions, but none are obvious. The [relevant bit of the lightkurve docs don't link elsewhere](https://docs.lightkurve.org/tutorials/1-getting-started/using-light-curve-file-products.html#3.-The-SAP-and-PDCSAP-Light-Curves). Eventually, let's have a brief expansion of the acronyms, and a literature reference for Kepler+TESS column definitions in our docs.

bmorris3 (approved):
I read everything and tested it and it works great, thanks! The comment I left doesn't block merge.

PR #68, opened 2023-12-01
jdaviz 3.9 updates

kecnry:
This PR encompasses minor updates for when we want to update the min-pin on jdaviz to 3.9

**Waiting for**: ~jdaviz 3.9~

- [x] revert pin of jdaviz main (and use >=3.9 instead)

Will also get from upstream:
* https://github.com/spacetelescope/jdaviz/pull/2640 
* https://github.com/spacetelescope/jdaviz/pull/2705
* https://github.com/spacetelescope/jdaviz/pull/2743
* https://github.com/spacetelescope/jdaviz/pull/2745

**TODO**:
- [x] revert [6a0086e](https://github.com/spacetelescope/lcviz/pull/78/commits/6a0086e8d6ce058097f946d44332d47ee9c36d46) once #78 is merged
- [x] include https://github.com/kecnry/lcviz/pull/3 if https://github.com/spacetelescope/jdaviz/pull/2733 is accepted upstream
- [x] include https://github.com/kecnry/lcviz/pull/4 if https://github.com/spacetelescope/jdaviz/pull/2742 is accepted upstream (and included in 3.9)
- [x] revert "delay overwriting existing registry item" from https://github.com/spacetelescope/lcviz/pull/94 if that is merged and https://github.com/spacetelescope/jdaviz/pull/2744 is accepted upstream
- [x] if  https://github.com/spacetelescope/lcviz/pull/94 is merged and https://github.com/spacetelescope/jdaviz/pull/2745 is accepted upstream, remove lcviz styling to prevent closing default time-viewer.
- [x] include https://github.com/kecnry/lcviz/pull/5 if https://github.com/spacetelescope/jdaviz/pull/2722 is accepted upstream
- [x] https://github.com/spacetelescope/lcviz/issues/92

PR #78, opened 2024-01-02
ephemeris component validation

kecnry:
This PR implements validation in ephemeris component naming to avoid breaking assumptions when parsing viewer names.

This won't do anything without the [upstream PR](https://github.com/spacetelescope/jdaviz/pull/2642), so the new tests have a condition to change behavior based on the version of jdaviz (which can then be removed in #68).

Eventually we might be able to remove this if/when we refactor the viewers to hold the information about their ephemeris, etc, within metadata instead of the label itself.  That would then allow us to set these labels by default, but allow the users to rename them without losing the ability to manage them from the plugins. [:cat:](https://jira.stsci.edu/browse/JDAT-4111)


kecnry:
I'm going to consider this simple enough and merge so I can get the follow-up into #68.  If anyone has concerns, we can make tweaks after.

PR #82, opened 2024-01-29
Feature: TPF support

kecnry:
currently a work in progress, not yet ready for merge

This feature branch contains:
* #75 
* #81 
* #83
* #85
* #86
* #101

**TODO**:
- [x] revert pin of jdaviz main (and use >=3.9 instead)

**Waiting for**: ~jdaviz 3.9~

PR #85, opened 2024-02-14
Time Selector (adapted version of cubeviz's slice) plugin

kecnry:
This PR implements a modified version of cubeviz's slice plugin into lcviz for TPF support.

NOTE: this PR requires the following upstream PRs and therefore an update to the jdaviz pin, and until then the RTD build and tests are expected to fail:
* https://github.com/spacetelescope/jdaviz/pull/2706
* https://github.com/spacetelescope/jdaviz/pull/2709
* https://github.com/spacetelescope/jdaviz/pull/2715

Now that these are all accepted and merged into jdaviz, this PR should be tested against jdaviz main (3.9).

And the following upstream draft PR was created for this, but is no longer used within this PR, so is not necessary:
* https://github.com/spacetelescope/jdaviz/pull/2711 (no longer used for the slice plugin - probably will be used once the photometric extraction plugin exists, though)

This will then also bump #82 to require the necessary pin of jdaviz (likely 3.9).

Example usage:

```
from lightkurve import search_lightcurve, search_targetpixelfile
from lcviz import LCviz

lc = search_lightcurve("HAT-P-11", mission="Kepler", cadence="long", quarter=10).download() #.flatten()
tpf = search_targetpixelfile("HAT-P-11", mission="Kepler", cadence="long", quarter=10).download()


lcviz = LCviz()
lcviz.load_data(lc)
lcviz.load_data(tpf)

po = lcviz.plugins['Plot Options']
po.viewer = 'image'
po.image_colormap = 'Gray'
po.stretch_function = 'Arcsinh'
po.stretch_vmin = 0
po.stretch_vmax = 20000

lcviz.show()
```

https://github.com/spacetelescope/lcviz/assets/877591/5753ca05-ada0-472b-9397-83839a33557c

https://github.com/spacetelescope/lcviz/assets/877591/8417cf12-31fb-4a6f-8f34-21707cab3a44


bmorris3:
It looks like the default selected tool is slice, even for light curves without TPFs. If you press Home, it keeps slice activated:

Screen Shot 2024-03-04 at 14 23 54

Is that intentional?

kecnry:
I _think_ what you're describing (if I followed it correctly) is a little more subtle (and quite a bit more complicated) than snapping vs not snapping.  There are a few coupled things going on here:
1. snap-to-slice currently means snap to the slice in a cube, not a data point in the viewer with the indicator.  Snapping is enabled here by default, but when there is _no_ cube, then it is effectively ignored.  I don't think we want to change this to "snap to data point in the indicator viewer", because then that can result in an annoying experience when a cube is loaded but sampled at different times than the light curve.  The main point of snapping is so that the indicator exactly matches the image shown.  We could have a separate option for "snap to point"... or when there is **no** cube loaded, snap-to-slice could then actually snap to data points (that feels inconsistent to me though and might be confusing logic).
2. even if you had a cube loaded, I don't think it would behave as you're expecting because the slice tool works only considers the slice (x) axis, whereas the mouseover display is doing nearest on the screen (in xy).  So should the mousover behavior change so that it feels more consistent with the slice tool?  Or does this different behavior make sense because they're different tools?  And if we want to make any changes, should those be isolated to lcviz or be made upstream?

kecnry:
> It looks like the default selected tool is slice, even for light curves without TPFs. If you press Home, it keeps slice activated.  Is that intentional?

Yes, I couldn't see any reason for it not to be the default since there currently was no other default.  Do you disagree with that?

kecnry:
This now supports dragging the indicator in the phase-viewer (by computing a delta time from a delta phase and setting the time in the same cycle).


https://github.com/spacetelescope/lcviz/assets/877591/0dea3706-e38e-4c05-b7f8-7bedd2f688b4

kecnry:
Ok, we're merging this as is into the feature branch and then will consider the points above as we flesh out the rest of the functionality.  Thanks for the feedback!

PR #91, opened 2024-02-23
Fix cloned phase-viewers when deleting/renaming an ephemeris component

kecnry:
Previously cloned phase-viewers (introduced in #74, not yet in a release) were not updated (renamed or removed) when an ephemeris component was renamed.  This PR updates the ephemeris plugin to be aware that there may be multiple phase-viewers corresponding to the ephemeris, and treats the "first" as the default when assigning defaults based on axes limits, etc.

(ignore the crashing time plot - too many notebooks running at once)

https://github.com/spacetelescope/lcviz/assets/877591/70eb135a-3562-49c9-9884-ed38e9e30991

devdeps failure is likely because of jdaviz, but we are still pinned to jdaviz 3.8 (and those tests are passing).  See #92 and #68.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.2.0 - unreleased
 ------------------
 
-* Clone viewer tool. [#74]
+* Clone viewer tool. [#74, #91]
 
 * Flux column plugin to choose which column is treated as the flux column for each dataset. [#77]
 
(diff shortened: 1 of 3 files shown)


rosteen (approved):
I tested the behavior on main and with this, definitely looks better now.

PR #94, opened 2024-03-06
Viewer creator in app toolbar

kecnry:
**NOTE**: this should be tested with jdaviz 3.8.x since this does not include updates for jdaviz 3.9 covered in #68.

This adds a viewer creator dropdown in the app-level toolbar which can create viewers, including phase viewers attached to a specific ephemeris.

https://github.com/spacetelescope/lcviz/assets/877591/c29ab69a-c3d6-4305-b860-d8ff484b864f

~The current behavior is to first check for an existing viewer with a matching label and if it exists, _clone_ the existing viewer (keeping data visibilities and plot options, etc), and otherwise create a new viewer.~

TODO:
- [x] testing with renaming ephemeris
- [x] either here or follow-up: move ephemeris logic into viewer metadata instead of relying on viewer label 
- [x] decide whether to allow all time viewers to be closed (and if so, test layer-visibility logic assumptions for new phase viewers)
- [x] upstream support for automatically opening data menu on blank viewer: https://github.com/spacetelescope/jdaviz/pull/2742 

Once merged:
- [ ] rebase the 3.9-updates and TPF feature branch on top of main and add support for image viewers (might require fixing clone support first)
- [ ] design and create ticket for what to show when no viewers exist?

The styling will also be improved by upstream https://github.com/spacetelescope/jdaviz/pull/2743

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,9 @@
-0.2.0 - unreleased
+0.3.0 - unreleased
+------------------
+
+* Ability to create additional viewers. [#94]
+
+0.2.0 (02-26-2024)
 ------------------
 
 * Clone viewer tool. [#74, #91]
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -7,10 +7,9 @@ from lightkurve import LightCurve
 from glue.core.component_id import ComponentID
 from glue.core.link_helpers import LinkSame
 from jdaviz.core.helpers import ConfigHelper
+from lcviz.viewers import TimeScatterView
 
 __all__ = ['LCviz']
-
-_default_time_viewer_reference_name = 'flux-vs-time'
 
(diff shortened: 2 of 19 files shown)


rosteen:
Functionality seems to work but I think the button really needs a tooltip - is there some difficulty with adding one to a downstream component like this?

rosteen:
> Yes, this will be covered by [spacetelescope/jdaviz#2743](https://github.com/spacetelescope/jdaviz/pull/2743) and [spacetelescope/jdaviz#2744](https://github.com/spacetelescope/jdaviz/pull/2744) upstream and pulled in here by #68.

Got it, thanks - I figured it wasn't trivial to insert something into out tooltip registry as-is. In that case I think this looks good to get in, I worry slightly about the number of things in the duplication menu getting out of hand, but also think it's very unlikely someone would create that many viewers.

PR #95, opened 2024-03-12
TST: See if percentile is breaking devdeps

pllim:
Attempt to understand failures in #93

PR #96, opened 2024-03-12
TST: See if percentile is breaking devdeps (try 2)

pllim:
Builds on top of #68 to investigate #93 . Supersedes #95 .


pllim:
`ERROR: Could not find a version that satisfies the requirement jdaviz>=3.9 (from versions: 0.1, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.1, 2.0.0, 2.1.0, 2.2.0, 2.3.0, 2.4.0, 2.5.0, 2.6.0, 2.7.0.post1, 2.7.1, 2.8.0, 2.9.0, 2.10.0, 3.0.0, 3.0.1, 3.0.2, 3.1.0, 3.1.1, 3.1.2, 3.2.0, 3.2.1, 3.3.0, 3.3.1, 3.4.0, 3.5.0, 3.6.0, 3.6.1, 3.6.2, 3.7.0, 3.7.1, 3.8.0, 3.8.1, 3.8.2)`

Wut...?

pllim:
Ignoring `KeyError: 'lcviz-viewer-creator'`, the percentile patch does not seem to help. Will continue to investigate.

lcviz/viewers.py
```diff
@@ -15,7 +15,8 @@
 from jdaviz.configs.default.plugins.viewers import JdavizViewerMixin
 from jdaviz.configs.specviz.plugins.viewers import SpecvizProfileView
 
-from lcviz.state import ScatterViewerState
+#from lcviz.state import ScatterViewerState
+from glue.viewers.scatter.state import ScatterViewerState
```
pllim, line 19: This is what actually fixed the failures, so I convinced myself now that the incompatibility is within `lcviz` and not `jdaviz`. However, in doing this, some results in `test_plugin_markers.py` changed and I am not sure why.

kecnry: we want the lcviz overrides for resetting limits though.  But something in there must not be playing nicely with the upstream PR in jdaviz that introduced the zoom-radius for Imviz 🤔

pllim: Yeah, pretty much. There is nothing to do over at Jdaviz for this.

kecnry: I am not 100% sure that is the case yet, we still need to understand why `ScatterViewerState` in lcviz is conflicting with the logic from https://github.com/spacetelescope/jdaviz/pull/2649.  The fix may be here or something may need to be generalized there more, I'm not sure yet.  But I think we can now be fairly certain that the configs in jdaviz are not affected

pllim: I am not sure if I am the best person to investigate more. I never worked on lcviz so I am not privy on what your scatter state customization is supposed to do.

kecnry:
Thanks for helping to track this down!  The underlying bug was fixed by https://github.com/kecnry/lcviz/pull/8 for jdaviz 3.9 and https://github.com/spacetelescope/lcviz/pull/102 for lcviz main (to avoid testing with jdaviz 3.9 in devdeps)

PR #101, opened 2024-03-27
enable clone viewer for image/TPF viewer

kecnry:
This pull request enables the clone viewer functionality for the TPF/image viewers as well as the ability to create a new blank cube viewer (so long as cube-like data exists in the data collection), and in doing so also:
* expands clone viewer logic to adopt layer-level plot options
* removes duplicated clone viewer logic

**NOTE**: this should be tested with jdaviz main (3.9) and is expected to fail tests on CI until rebased on top of #68.  Tests do pass locally for me and we'll ensure they pass before merging #82 into main.

https://github.com/spacetelescope/lcviz/assets/877591/1bc25754-4260-4af7-857a-3743e5c6c458

### DIFF
--- a/lcviz/plugins/viewer_creator/viewer_creator.py
+++ b/lcviz/plugins/viewer_creator/viewer_creator.py
@@ -40,7 +40,7 @@ class ViewerCreator(ViewerCreator):
 
         self.viewer_types = [v for v in self.viewer_types if v['name'].startswith('lcviz')
                              and not v['label'].startswith('flux-vs-phase')
-                             and not v['label'] == 'cube'] + phase_viewers + cube_viewers
+                             and not v['label'] in ('cube', 'image')] + phase_viewers + cube_viewers
         self.send_state('viewer_types')
 
     def vue_create_viewer(self, name):


lcviz/plugins/viewer_creator/viewer_creator.py
```diff
@@ -24,13 +29,18 @@ def _rebuild_available_viewers(self, *args):
         if self.app._jdaviz_helper is not None:
             phase_viewers = [{'name': f'lcviz-phase-viewer:{e}', 'label': f'flux-vs-phase:{e}'}
                               for e in self.app._jdaviz_helper.plugins['Ephemeris'].component.choices]  # noqa
+            if self.app._jdaviz_helper._has_cube_data:
```
kecnry, line 32: right now this enables creating an "image" viewer if there is any cube-like data in the app.  In the future, we may want to change this to allow image-like data, but that is not yet supported at the parser level.

lcviz/viewers.py
```diff
@@ -253,33 +259,6 @@ def apply_roi(self, roi, use_current=False):
...
-        new_viewer = self.jdaviz_app.get_viewer(name)
-        if hasattr(self, 'ephemeris_component'):
-            new_viewer._ephemeris_component = self._ephemeris_component
-        for k, v in this_state.items():
-            if k in ('layers',):
-                continue
-            setattr(new_viewer.state, k, v)
-
-        return new_viewer.user_api
-
```
kecnry, line 282: this was duplicated logic that was somehow introduced when refactoring to use the mixin for clone viewer support

bmorris3 (approved):
This looks good. I know this comment isn't specific to this PR, but what is the viewer icon meant to represent? Is it a histogram?

kecnry:
> but what is the viewer icon meant to represent? Is it a histogram?

yeeeeaaah, I think its just supposed to generically represent "graph" or "plot".  I agree it isn't ideal, but that is defined in jdaviz and adopted here 🤷‍♂️

PR #102, opened 2024-03-28
temporarily remove jdaviz from devdeps

kecnry:
We're currently pinning a minor version of jdaviz and handling updates manually, so want to test that version of jdaviz with other devdeps (so a break caused by jdaviz does not prevent us from testing the other devdeps).  Updating the pin to the next minor version of jdaviz is currently handled by PRs that bump the version and apply any necessary changes (i.e. #68) which will then ensure tests pass before bumping the minor version of jdaviz.

If we ever stop tagging a specific minor version, dev tests for jdaviz should be re-introduced (as mentioned in the added comment in `pyproject.toml`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,8 @@ classifiers = [
 ]
 dependencies = [
     "astropy>=5.2",
+    # NOTE: if/when we stop pinning a minor version of jdaviz, add jdaviz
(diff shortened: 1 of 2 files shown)


rosteen (approved):

rosteen:
Seems like a good plan for now.

PR #104, opened 2024-04-12
TPF viewer: marker/mouseover support

kecnry:
This PR adds full mouseover information and marker plugin support for the TPF viewer.

### DIFF
--- a/lcviz/plugins/coords_info/coords_info.py
+++ b/lcviz/plugins/coords_info/coords_info.py
@@ -52,7 +52,8 @@ class CoordsInfo(CoordsInfo):
             self._dict['data_label'] = ''
             self._dict['time'] = x if not is_phase else np.nan
             self._dict['phase'] = x if is_phase else np.nan
-            self._dict['flux'] = y
+            self._dict['value'] = y
+            self._dict['value:unit'] = y_unit
             self._dict['ephemeris'] = ''
 
             self.row2_title = ''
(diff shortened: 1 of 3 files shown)


lcviz/plugins/coords_info/coords_info.py
```diff
@@ -52,7 +52,8 @@ def _cursor_fallback():
             self._dict['data_label'] = ''
             self._dict['time'] = x if not is_phase else np.nan
```
bmorris3, line 53: Can we add a time unit?

lcviz/plugins/coords_info/coords_info.py
```diff
@@ -139,13 +140,92 @@ def _cursor_fallback():
...
+            self._dict['data_label'] = image.label
+
+        if self.dataset.selected == 'none' or image is None:
+            self.icon = 'mdi-cursor-default'
+            self._dict['data_label'] = ''
+        else:
+            time = viewer.slice_value
+            self.row2_title = 'Time'
+            self.row2_text = f'{time:0.5f}'
+            self._dict['time'] = time
```
bmorris3, line 197: Units please?

bmorris3 (approved):
This works great for me with jdaviz==3.9.0.

One important thing: when you run
```python
markers = lcviz.plugins['Markers']
markers.export_table()
```
you get a table with the times in the `dt` units. Let's transform those to proper JD/MJD before export:

Screen Shot 2024-04-19 at 11 07 57

bmorris3:
And to be clear – I have approved so we can merge once we add units or decide to do that later,  fix the time column in the exported marker table, and tests pass.

kecnry:
Added the time unit - good catch. 

I'm a bit conflicted on the handling of reference time to get back to JD, etc.  If the x-axes and mouseover show `dt`, then I think it makes sense for that to be maintained during export.  But maybe we could/should include the reference time in the table as well?  Or maybe we just really need to think about reference time in general and have its behavior configured in a plugin (this is on our roadmap, but we haven't really fleshed out what that would look like yet).  Thoughts?  Maybe we should defer that for now and link back to this discussion for when we take on that effort?

PR #106, opened 2024-04-16
raise error when parser can't identify file_obj

kecnry:
rather than an internal error that light_curve is not defined.

To reproduce:

```
from lcviz import LCviz
lcviz = LCviz()
lcviz.load_data(None)
```


bmorris3 (approved):
Thanks, I actually needed this _today_.

PR #107, opened 2024-04-16
Stitch plugin

kecnry:
https://github.com/spacetelescope/lcviz/assets/877591/e2a8b571-e8ab-41b2-a163-162eaf0f719a

[Stitch plugins docs](https://lcviz--107.org.readthedocs.build/en/107/plugins.html#stitch)

Potential follow-up:
* normalize/flatten capabilities or custom user-provided `corrector_func`. 
 See https://docs.lightkurve.org/reference/api/lightkurve.LightCurveCollection.stitch.html
 
 Open questions:
- [x] should there be functionality to either unload the input entries from the viewer or delete from the app entirely _within_ this plugin (so effectively replace the individual entries with the stitched entry)?
 - [x] should this be called "stitch", "combine", "merge", etc?

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -213,6 +213,47 @@ The time selector plugin allows defining the time indicated in all light curve v
 
 
 
+.. _stitch:
+
+Stitch
+======
+
+This plugin allows for combining multiple light curves into a single entry.  Note that this plugin
+is only available if there are at least two light curves loaded into a light curve viewer.
+
+.. admonition:: User API Example
+    :class: dropdown
+
+    See the :class:`~lcviz.plugins.stitch.stitch.Stitch` user API documentation for more details.
+
+    .. code-block:: python
+
+      from lcviz import LCviz
+      lc1 = search_lightcurve("HAT-P-11", mission="Kepler",
+                              cadence="long", quarter=9).download()
+      lc2 = search_lightcurve("HAT-P-11", mission="Kepler",
+                              cadence="long", quarter=10).download()
+      lcviz = LCviz()
+      lcviz.load_data(lc1, 'lc1')
+      lcviz.load_data(lc2, 'lc2')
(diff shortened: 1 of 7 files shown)


bmorris3:
Some thoughts:
* "Stitch" is the word used for this operation in lightkurve, let's use that too
* I like the idea of optionally deleting data entries after stitch. there could be a toggle in the plugin set to "delete after stitch" by default.
* in most cases, stitching without applying a corrector function is not useful. this would be the first place in lcviz where we could have a Model Fitting-like interface outside of jdaviz – we could support specifying a functional form for the baseline and fitting it, or taking simple statistics (mean/median/rolling mean or median) of each segment.
* it looks like we need to update the data labeler to put in the quarter/sector (should we call these "segments"?) info.

kecnry:
> "Stitch" is the word used for this operation in lightkurve, let's use that too

✅ 🪡 

> I like the idea of optionally deleting data entries after stitch. there could be a toggle in the plugin set to "delete after stitch" by default.

ok, then I'll add it (but not sure about having it on by default - we'll see as we play with it I guess)

> in most cases, stitching without applying a corrector function is not useful.

I'm not sure I agree with this.  All of the other plugins take a single light curve as input, so this allows you to combine multiple quarters/sectors together and then use that as input to... frequency analysis, binning, etc.  Down the road, we may want to have that done on the fly by allowing any plugin to have multi-select support for the dataset and stitch on the fly, but again, without necessarily normalizing or flattening as part of that process.

> this would be the first place in lcviz where we could have a Model Fitting-like interface outside of jdaviz – we could support specifying a functional form for the baseline and fitting it, or taking simple statistics (mean/median/rolling mean or median) of each segment.

Maybe, but definitely not for this first pass.  And I think we want model fitting on its own... it just needs to be completely refactored to not rely on specutils.

> it looks like we need to update the data labeler to put in the quarter/sector (should we call these "segments"?) info.

Good idea - I support this.

PR #108, opened 2024-04-17
metadata: monkeypatch Undefined to show as empty string

### LINKED ISSUE #88: some metadata display as python pointer rather than value
Some of the metadata keys display values as pointer objects rather than values.  For example, with https://masttest.stsci.edu/viz/ui/#/lcviz?uri=mast%3AKEPLER%2Furl%2Fmissions%2Fkepler%2Flightcurves%2F0107%2F010748390%2Fkplr010748390-2009131110544_slc.fits

Click on Metadata viewer.  These keys display the following.  
```
CDPP12_0  <astropy.io.fits.card.Undefined object at 0x7f97b01acd90>
CDPP3_0   <astropy.io.fits.card.Undefined object at 0x7f97b01acd90>
CDPP6_0   <astropy.io.fits.card.Undefined object at 0x7f97b01acd90>
```

kecnry:
This PR closes #88 by monkeypatching the string representation of `astropy.io.fits.card.Undefined` to show an empty string in the metadata plugin instead of something like `<astropy.io.fits.card.Undefined object at 0x29f5b94d0>`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ------------------
 
 * Support loading, viewing, and slicing through TPF data cubes. [#82]
+* Metdata plugin: show undefined entries as empty string instead of object repr. [#108]
 
 0.3.0 - (04-05-2024)
 --------------------
--- a/lcviz/plugins/metadata_viewer/metadata_viewer.py
+++ b/lcviz/plugins/metadata_viewer/metadata_viewer.py
@@ -1,7 +1,14 @@
+from astropy.io.fits.card import Undefined
+
 from jdaviz.configs.default.plugins import MetadataViewer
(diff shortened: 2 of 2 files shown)


bmorris3 (approved):
For lcviz, and for now, this is fine. Let's fix this upstream in jdaviz though, so we don't have to rely on a monkeypatch on an astropy class.

kecnry:
I did first look into a jdaviz fix, but it would require probably introducing inefficiencies to this [map line](https://github.com/spacetelescope/jdaviz/blob/main/jdaviz/configs/default/plugins/metadata_viewer/metadata_viewer.py#L122) by adding if-statements to override this particular case.  That may be worth it at some point though if there are other situations that need special treatment as well 🤔

PR #111, opened 2024-04-19
replace flux-origin with quarter/campaign/sector in data_label

kecnry:
This PR modifies the default data_labels for light curves to _remove_ flux-origin (since that can now be changed later by the flux origin plugin) and _adds_ quarter/campaign/sector information.

Note that calling `stitch` before loading into lcviz, will result in the _last_ quarter/campaign/sector being adopted in the metadata and therefore into the data-label.  Is this something we should propose for changes upstream?

Also note: TPF data are still suffixed with `[TPF]` in this PR, although that could be changed now or in the future if desired.

Discussion points:
- [ ] syntax for the suffixes (currently is C/Q/S followed by the number)

### DIFF
--- a/lcviz/parsers.py
+++ b/lcviz/parsers.py
@@ -37,11 +37,16 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
         new_data_label = light_curve.meta.get('OBJECT', 'Light curve')
 
     # handle flux_origin default
+    mission = light_curve.meta.get('MISSION', '')
     flux_origin = light_curve.meta.get('FLUX_ORIGIN', None)  # i.e. PDCSAP or SAP
     if isinstance(light_curve, lightkurve.targetpixelfile.TargetPixelFile):
         new_data_label += '[TPF]'
-    elif flux_origin is not None:
-        new_data_label += f'[{flux_origin}]'
+    elif mission == 'Kepler':
+        new_data_label += f' Q{light_curve.meta.get("QUARTER")}'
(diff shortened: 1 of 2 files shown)


lcviz/parsers.py
```diff
@@ -37,11 +37,16 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
         new_data_label = light_curve.meta.get('OBJECT', 'Light curve')
 
     # handle flux_origin default
+    mission = light_curve.meta.get('MISSION', '')
```
bmorris3, line 40:
```suggestion
    mission = light_curve.meta.get('MISSION', '').lower()
```

lcviz/parsers.py
```diff
@@ -37,11 +37,16 @@ def light_curve_parser(app, file_obj, data_label=None, show_in_viewer=True, **kw
...
     flux_origin = light_curve.meta.get('FLUX_ORIGIN', None)  # i.e. PDCSAP or SAP
     if isinstance(light_curve, lightkurve.targetpixelfile.TargetPixelFile):
         new_data_label += '[TPF]'
-    elif flux_origin is not None:
-        new_data_label += f'[{flux_origin}]'
+    elif mission == 'Kepler':
```
bmorris3, line 44:
```suggestion
    elif mission == 'kepler':
```

bmorris3 (approved):
Let's just make sure capitalization doesn't matter. Otherwise looks good.

PR #117, opened 2024-04-26
Better default TPF stretch

bmorris3:
This PR gives a better default colormap stretch on TPF files. It sets vmin/vmax to the 5/99 percentile limits from the cube:

```python
from lightkurve import search_targetpixelfile, search_lightcurve
from lcviz import LCviz

target = 'TRAPPIST-1'
lc = search_lightcurve(target, mission="K2", cadence="long").download().normalize()
tpf = search_targetpixelfile(target, mission="K2", cadence="long").download()

# Load the light curve into LCviz:
lcviz = LCviz()
lcviz.load_data(lc)
lcviz.load_data(tpf)
lcviz.show()

viewer = lcviz.app.get_viewer('flux-vs-time')
viewer.state.y_min = 0.9
viewer.state.y_max = 1.2
```

Screen Shot 2024-04-26 at 12 45 46

I opted to put this stretch as a private method in `PlotOptions`, and call the method from the parser when a TPF is loaded. That keeps the diff small. Caveats: this approach won't work if we load more than one TPF into the `image` viewer.

### DIFF
--- a/lcviz/plugins/plot_options/plot_options.py
+++ b/lcviz/plugins/plot_options/plot_options.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from jdaviz.configs.default.plugins import PlotOptions
 from jdaviz.core.registries import tray_registry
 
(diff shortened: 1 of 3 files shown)


lcviz/plugins/plot_options/plot_options.py
```diff
@@ -31,6 +33,20 @@ def __init__(self, *args, **kwargs):
...
+        viewer = self.app.get_viewer(tpf_viewer_reference)
+        image = viewer.layers[0].get_image_data()
+        vmin, vmax = np.nanpercentile(
+            image, [vmin_percentile, vmax_percentile]
+        )
+
+        self.viewer_selected = tpf_viewer_reference
+        self.stretch_function_value = 'log'
+        self.stretch_vmin_value = vmin
+        self.stretch_vmax_value = vmax
```
kecnry, line 48: can we do this in a separate copy of the plugin to avoid changing the UI itself?

```suggestion
        po = self.new()
        po.viewer_selected = tpf_viewer_reference
        po.stretch_function_value = 'log'
        po.stretch_vmin_value = vmin
        po.stretch_vmax_value = vmax
```

kecnry (commented):
> I opted to put this stretch as a private method in PlotOptions, and call the method from the parser when a TPF is loaded. That keeps the diff small. Caveats: this approach won't work if we load more than one TPF into the image viewer.

What if we listened to adding new data or layers (either here or in the helper... but I would prefer in the plugin I think to keep the code in one place)?

kecnry (approved):
Thanks!!

PR #120, opened 2024-06-18
Prevent duplicate sub-interval data labels

bmorris3:
If you load Kepler data from Quarter 10 with a `data_label = "Target in Q10"`, the parser will happily add another "Q10" to the end of the data label [🐱](https://jira.stsci.edu/browse/JDAT-4565). This PR checks if the sub-interval for any mission is already in the data label, and only adds one if none is found.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     # to devdeps in tox.ini
     "jdaviz>=3.10.2,<3.11",
     "lightkurve>=2.4.1",
+    "numpy<2",
 ]
 dynamic = [
     "version",
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.5.0 (unreleased)
 ------------------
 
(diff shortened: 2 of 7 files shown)


bmorris3:
It looks like lighkurve is calling functions in [oktopus](https://github.com/KeplerGO/oktopus), which depends on [autograd](https://github.com/HIPS/autograd), which is no longer maintained. This was causing [a test failure](https://github.com/spacetelescope/lcviz/actions/runs/9571863556/job/26389888439).

kecnry:
The other option would be to just not append at all if manually passing a data label - can you think of a case where providing a data label would still want the quarter/whatever automatically appended?

kecnry (approved):
Thanks!

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     # to devdeps in tox.ini
     "jdaviz>=3.10.2,<3.11",
     "lightkurve>=2.4.1",
+    "numpy<2",
```
kecnry, line 27: backported in #126

PR #122, opened 2024-06-26
fix CDIPS parsing support (column filled with strings)

kecnry:
which affected CDIPS HLSP light curves and prevented them from being loaded into lcviz.  Note that the time format is still incorrect, but that is a separate issue.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 0.4.1 (unreleased)
 ------------------
 
+* Fixes CDIPS support by handling units stored as empty strings. [#122]
+
 0.4.0 (06-11-2024)
 ------------------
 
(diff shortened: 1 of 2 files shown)


bmorris3 (approved):
Simple enough fix, and I checked that it works for me by running:
```python
from lcviz import LCviz
from astroquery.mast import Observations

ticid = '210400528'
obs_table = Observations.query_criteria(target_name=ticid, provenance_name="CDIPS")
print("Found " + str(len(obs_table)) + " CDIPS light curves.")

# Get list of available products for this Observation.
cdips_products = Observations.get_product_list(obs_table)

manifest = Observations.download_products(cdips_products)

lcviz = LCviz()

for path in manifest['Local Path']:
    lcviz.load_data(path)

lcviz.show()
```

It raises lots of warnings about the units of many columns, but I'm pretty sure that's on the astropy Tables+this data product, and not on LCviz.

PR #126, opened 2024-07-15
(temporarily) max pin numpy<2

kecnry:
Was included in #120, but never backported.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 0.4.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


dhomeier:
@kecnry I have pushed lightkurve/lightkurve#1452 to remove the strict dependency on oktopus; in a fresh environment this allows installation with numpy 2.0 for Python 3.10+, although in the CI still some other packages are blocking numpy > 1.24.

dhomeier:
Fixed in lightkurve/main now (for 2.5.1)

PR #127, opened 2024-07-16
Add methods for querying NASA Exoplanet Archive for ephemeris

bmorris3:
### Demo

```python
from lcviz import LCviz

lc = search_lightcurve(
    "TRAPPIST-1", mission='K2', author='EVEREST'
)[0].download().remove_nans().normalize().remove_outliers()

# load into LCviz:
lcviz = LCviz()
lcviz.load_data(lc)
lcviz.show()

# flatten the light curve to remove rotational modulation:
flatten = lcviz.plugins['Flatten']
flatten.window_length = 31
flatten.flatten();

# query for the planets in this system:
ephem = lcviz.plugins['Ephemeris']
ephem.query_for_ephemeris()

# make a new ephemeris viewer for the first three planets:
for planet in ephem.query_result.choices[:3]:
    ephem.query_result = planet
    ephem.adopt_from_catalog_in_new_viewer()
```

Screen Shot 2024-07-22 at 16 29 52

### DIFF
--- a/lcviz/plugins/ephemeris/ephemeris.py
+++ b/lcviz/plugins/ephemeris/ephemeris.py
@@ -700,12 +700,12 @@ class Ephemeris(PluginTemplateMixin, DatasetSelectMixin):
 
     def adopt_from_catalog_in_new_viewer(self, *args):
         new_component_label = self.query_result_selected.replace(' ', '')
-        if len(self._get_phase_viewers(new_component_label)):
+        if new_component_label in self.component.choices:
             # warn the user that an ephemeris component already exists with this label,
             # a second won't be added:
             self.hub.broadcast(
                 SnackbarMessage(
-                    f"Ephemeris component {new_component_label} already exists, skipping",
+                    f"Ephemeris component {new_component_label} already exists, skipping.",
                     sender=self, color="warning"
                 )
             )
--- a/lcviz/plugins/ephemeris/ephemeris.py
+++ b/lcviz/plugins/ephemeris/ephemeris.py
@@ -15,6 +15,7 @@ from jdaviz.core.template_mixin import (PluginTemplateMixin, DatasetSelectMixin,
                                         SelectPluginComponent, EditableSelectPluginComponent,
                                         with_spinner)
 from jdaviz.core.user_api import PluginUserApi
+from jdaviz.core.events import SnackbarMessage
 
 from lightkurve import periodogram, FoldedLightCurve
 
@@ -137,6 +138,10 @@ class Ephemeris(PluginTemplateMixin, DatasetSelectMixin):
                                             selected='method_selected',
                                             manual_options=['Lomb-Scargle', 'Box Least Squares'])
 
+        self.query_result = SelectPluginComponent(self,
+                                                  items='query_result_items',
+                                                  selected='query_result_selected')
+
         # TODO: could optimize by only updating for the new data entry only
         # (would require some refactoring and probably wouldn't have significant gains)
         self.hub.subscribe(self, DataCollectionAddMessage, handler=self._update_all_phase_arrays)
@@ -153,7 +158,7 @@ class Ephemeris(PluginTemplateMixin, DatasetSelectMixin):
             'times_to_phases', 'phases_to_times', 'get_data',
             'dataset', 'method', 'period_at_max_power',
             'adopt_period_at_max_power', 'query_for_ephemeris',
-            'query_result'
+            'query_result', 'adopt_from_catalog', 'adopt_from_catalog_in_new_viewer'
         ]
         return PluginUserApi(self, expose=expose)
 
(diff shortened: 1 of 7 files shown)


lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -578,3 +597,66 @@ def get_data(self, dataset, ephem_component=None):
...
+        elif None not in (self.query_ra, self.query_dec):
+            coord = SkyCoord(ra=self.query_ra, dec=self.query_dec, unit=u.deg)
+            self._query_result = self.nasa_exoplanet_archive.query_region(
+                coord, self.query_radius * u.arcsec,
+                table='pscomppars'
+            )
+
+        else:
+            # no metadata found for RA, Dec, or object name
+            return None
```
kecnry, line 629: can we either have the UI reflect this logic (only show coordinate inputs if the name entry is empty or something) or just require inputing a name (is there ever a case where someone would actually want to use this by coordinates?  I guess maybe to search for background contamination?)?

kecnry (commented):
I like it!  Just a few suggestions, mostly so that we can move towards exposing the API.  This will also need test coverage before merge (but I'm guessing that's partly why its still in draft).

I'm not too worried yet about whether this will ultimately live here or need to be shared by other plugins - we're before 1.0 release and haven't committed to backwards compatibility yet, so let's just implement things and we can always change them.

lcviz/plugins/ephemeris/ephemeris.py
kecnry: That's fine, I just think we need to think of a UI/UX that helps make that clear.  So either a dropdown to choose whether to query by identifier or coords, or have the coords hide when the identifier is not blank might help to make that feel intuitive without having to look in the docs.

lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -578,3 +597,66 @@ def get_data(self, dataset, ephem_component=None):
...
+
+        if len(self._query_result):
+            self._query_result.add_index('pl_name')
+            self.query_result_names = sorted(list(self._query_result['pl_name']))
+
+    @observe('query_result_selected')
+    def _select_query_result(self, *args, selected_idx=None):
+        selected_query_result = self._query_result.loc[self.query_result_selected]
+        self.query_ra = selected_query_result['ra'].base.value
+        self.query_dec = selected_query_result['dec'].base.value
```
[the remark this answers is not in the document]
bmorris3: Agreed, removing.

lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -578,3 +607,109 @@ def get_data(self, dataset, ephem_component=None):
...
+            self.hub.broadcast(
+                SnackbarMessage(
+                    f"Ephemeris component {new_component_label} already exists, skipping.",
+                    sender=self, color="warning"
+                )
+            )
+        else:
+            self.add_component(new_component_label)
+            self.create_phase_viewer()
+            self.adopt_from_catalog()
```
kecnry, line 712: do we still need both of these now that there is only one button?  Maybe just one that is `add_component_from_query` (or similar)?  If we want to keep both, I think we need to use `new_component` instead of `viewer` since a viewer may or may not exist for an existing ephemeris component.

kecnry (commented):
I think eventually we may want to move this up or even into the ephemeris component... component, but this is good for now until we decide what else we want to support for querying (other catalogs, loading from metadata, etc).

Just one comment on the diff and then we can always iterate further later before an official 1.0 release - this definitely is useful to get in ASAP!

bmorris3:
@kecnry I've renamed `adopt_from_catalog_in_new_viewer` to `create_ephemeris_from_query`, and removed the unused method `adopt_from_catalog`.

lcviz/plugins/ephemeris/ephemeris.py
```diff
@@ -125,12 +150,16 @@ def __init__(self, *args, **kwargs):
...
-                  'dataset', 'method', 'period_at_max_power', 'adopt_period_at_max_power']
+        expose = [
+            'component', 'period', 'dpdt', 't0', 'wrap_at',
+            'ephemeris', 'ephemerides',
+            'update_ephemeris', 'create_phase_viewer',
+            'add_component', 'remove_component', 'rename_component',
+            'times_to_phases', 'phases_to_times', 'get_data',
+            'dataset', 'method', 'period_at_max_power',
+            'adopt_period_at_max_power', 'query_for_ephemeris',
+            'query_result', 'adopt_from_catalog', 'create_ephemeris_from_query'
```
kecnry, line 161:
```suggestion
            'query_result', 'create_ephemeris_from_query'
```

kecnry, line 161: `query_for_ephemeris`, `query_result`, and `create_ephemeris_from_query` need to be added to the API docs.

kecnry (commented):
Functionality looks good to me now, just a few requests for API docs, but otherwise I think this is good to go in, thanks!

PR #128, opened 2024-07-17
Use lcviz version in all docs links

kecnry:
This PR ensures that all internal docs links buttons point to the correct version on RTD, and don't use the upstream version tag set by jdaviz.

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -9,6 +9,8 @@ from glue.core.component_id import ComponentID
 from glue.core.link_helpers import LinkSame
 from glue.core.units import unit_converter
 from jdaviz.core.helpers import ConfigHelper
+
+from lcviz import __version__
 from lcviz.viewers import TimeScatterView
 
 __all__ = ['LCviz']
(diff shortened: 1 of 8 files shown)


lcviz/helper.py
```diff
@@ -134,7 +136,13 @@ def __init__(self, *args, **kwargs):
         self.app._add_style((__file__, 'lcviz_style.vue'))
 
         # set the link to read the docs
-        self.app.docs_link = "https://lcviz.readthedocs.io"
+        self.app.vdocs = 'latest' if 'dev' in __version__ else 'v'+__version__
```
kecnry, line 139: to test, override this with `'v0.4.1'` and see that all links in plugins and the app toolbar point correctly to that tagged release on RTD.

bmorris3:
Is this only necessary for RTD? If so, can it be solved with the same approach as https://github.com/spacetelescope/jdaviz/pull/3080?

kecnry:
this is links _from_ the app to RTD (like the button in the app-toolbar and "learn more" links in each plugin - so no).

bmorris3 (approved):
This works for me locally. I regret that any of this needs to be accomplished by observing a traitlet, but it's not clear how to avoid that without something much more complicated.

PR #137, opened 2024-08-30
improve initial scatter/mouseover performance

kecnry:
This along with https://github.com/glue-viz/glue-jupyter/pull/467 (included in glue-jupyter 0.22.2) improves the performance at load time, especially when the server is remote.

Note that lcviz does not explicitly pin glue-jupyter (it pulls in glue-jupyter through jdaviz).

### DIFF
--- a/lcviz/plugins/coords_info/coords_info.py
+++ b/lcviz/plugins/coords_info/coords_info.py
@@ -89,10 +89,7 @@ class CoordsInfo(CoordsInfo):
             if self.dataset.selected != 'auto' and self.dataset.selected != lyr.layer.label:
                 continue
 
-            # glue-jupyter 1.18 changed from lyr.scatter to lyr.scatter_mark
-            # TODO: once glue-jupyter is pinned to 1.18 or later, update this to:
-            # scatter = lyr.scatter_mark
-            scatter = getattr(lyr, 'scatter_mark', getattr(lyr, 'scatter', None))
+            scatter = lyr.scatter_mark
             lyr_x, lyr_y = scatter.x, scatter.y
             if not len(lyr_x):
                 continue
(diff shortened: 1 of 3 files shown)


lcviz/viewers.py
```diff
@@ -149,6 +149,7 @@ def _apply_layer_defaults(self, layer_state):
         if getattr(layer_state.layer, 'meta', {}).get('Plugin', None) == 'Binning':
             # increased size of binned results, by default
             layer_state.size = 5
+        layer_state.points_mode = 'markers'
```
kecnry, line 152: this isn't needed in order to get the improvements from glue-jupyter, but doesn't hurt!  If we ever have light curves with enough points where the scatter with markers becomes more expensive than the density plot, we can consider removing this to leave it at the default of 'auto'.

lcviz/plugins/coords_info/coords_info.py
```diff
@@ -89,10 +89,7 @@ def _cursor_fallback():
             if self.dataset.selected != 'auto' and self.dataset.selected != lyr.layer.label:
                 continue
 
-            # glue-jupyter 1.18 changed from lyr.scatter to lyr.scatter_mark
-            # TODO: once glue-jupyter is pinned to 1.18 or later, update this to:
-            # scatter = lyr.scatter_mark
-            scatter = getattr(lyr, 'scatter_mark', getattr(lyr, 'scatter', None))
+            scatter = lyr.scatter_mark
```
kecnry, line 92: lcviz currently requires jdaviz 3.10, which in turn requires glue-jupyter 0.20+ (original comment was a typo glue-jupyter made this change in 0.18, not 1.18, with https://github.com/glue-viz/glue-jupyter/pull/363)

PR #156, opened 2024-12-02
Ignore mpl_scatter_density warning temporarily

bmorris3:
Since https://github.com/spacetelescope/jdaviz/pull/3317, `jdaviz` ignores a warning raised by `mpl_scatter_density` (https://github.com/astrofrog/mpl-scatter-density/issues/46, https://github.com/astrofrog/mpl-scatter-density/pull/47). We should use the same ignore in lcviz until the upstream fix is released.


kecnry (approved):
Thanks!

PR #158, opened 2024-12-04
jdaviz 4.1 updates

kecnry:
* rename subset tools plugin (https://github.com/spacetelescope/jdaviz/pull/3293 and https://github.com/spacetelescope/jdaviz/pull/3304)

**Waiting for**: jdaviz 4.1

### DIFF
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -104,7 +104,7 @@ class LCviz(ConfigHelper):
                      'context': {'notebook': {'max_height': '600px'}}},
         'toolbar': ['g-data-tools', 'g-subset-tools', 'g-viewer-creator', 'lcviz-coords-info'],
         'tray': ['lcviz-metadata-viewer', 'flux-column',
-                 'lcviz-plot-options', 'lcviz-subset-plugin',
+                 'lcviz-plot-options', 'lcviz-subset-tools',
                  'lcviz-markers', 'time-selector',
                  'stitch', 'flatten', 'frequency-analysis', 'ephemeris',
                  'binning', 'lcviz-export'],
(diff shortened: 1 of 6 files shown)


kecnry:
(will need to track down the cause for failing checks - I suspect they're unrelated to this PR and might be due to a change in remote data)

kecnry:
These test result changes likely came from https://github.com/spacetelescope/jdaviz/pull/3235

We need to track down the cause for the changes in tests and if they are:
* fixes upstream and therefore expected changes
* needs an additional fix to jdaviz
* needs fixes here to account for changes in jdaviz

Note that the changes are _not_ just caused by changes to the remote data, as the same changes put on top of main are not necessary (see #173)

lcviz/tests/test_plugin_markers.py
```diff
@@ -147,18 +147,18 @@ def test_tpf_markers(helper, light_curve_like_kepler_quarter):
                                         {'event': 'mousemove',
                                          'domain': {'x': 0, 'y': 0}})
 
-    assert label_mouseover.as_text() == ('Pixel x=00000.0 y=00000.0 Value +1.27220e+00 electron / s',  # noqa
-                                         'Time 47.00000 d',
+    assert label_mouseover.as_text() == ('Pixel x=00000.0 y=00000.0 Value +1.28035e+01 electron / s',  # noqa
+                                         'Time 47.00689 d',
                                          '')
 
     print(label_mouseover.as_dict())
```
bmorris3:
```suggestion
```

kecnry: (reverted the changes to the tests - including this debug line - since #173 proved they were not caused by changes to upstream data)

kecnry:
remaining dev-test failures are (likely) related to https://github.com/astropy/astropy/issues/17859 and https://github.com/lightkurve/lightkurve/issues/1485

PR #164, opened 2024-12-31
Add parser for TESS DVT files

rosteen:
By default, this will load each TCE extension into the data collection as a separate object and display them all. Specifying `extname` will skip adding the deselected extensions into viewers. Opening as draft since I need to fix the auto-generated data label. I probably also need to think a little harder about what to do with the ephemeris in the case of multiple TCE extensions that are all added to the viewer.

Screenshot 2024-12-31 at 12 49 10 PM

### DIFF
--- a/lcviz/components/components.py
+++ b/lcviz/components/components.py
@@ -1,6 +1,5 @@
 from astropy import units as u
 from functools import cached_property
-import numpy as np
 
 from ipyvuetify import VuetifyTemplate
 from glue.core import HubListener
--- a/lcviz/components/components.py
+++ b/lcviz/components/components.py
@@ -223,7 +223,7 @@ class FluxColumnSelect(SelectPluginComponent):
         # manipulate the arrays in the data-collection directly, and modify FLUX_ORIGIN so that
         # exporting back to a lightkurve object works as expected
         self.app._jdaviz_helper._set_data_component(dc_item, 'flux', dc_item[self.selected])
-        if self.selected+"_err" in [c.label for c in dc_item.components]:
+        if self.selected+"_err" in dc_item.component_ids():
             self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err',
                                                         dc_item[self.selected+"_err"])  # noqa
         else:
--- a/lcviz/components/components.py
+++ b/lcviz/components/components.py
@@ -224,12 +224,13 @@ class FluxColumnSelect(SelectPluginComponent):
         # exporting back to a lightkurve object works as expected
         self.app._jdaviz_helper._set_data_component(dc_item, 'flux', dc_item[self.selected])
         if self.selected+"_err" in dc_item.component_ids():
-            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err',
-                                                        dc_item[self.selected+"_err"])  # noqa
+            if "flux_err" in dc_item.component_ids():
+                self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err',
+                                                            dc_item[self.selected + "_err"])
+            else:
+                dc_item.add_component(dc_item[self.selected + "_err"], 'flux_err')
         else:
-            nan_errs = np.empty(dc_item['flux'].shape)
-            nan_errs[:] = np.nan
-            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err', nan_errs)
+            dc_item.remove_component(dc_item.find_component_id('flux_err'))
 
         dc_item.meta['FLUX_ORIGIN'] = self.selected
 
--- a/lcviz/components/components.py
+++ b/lcviz/components/components.py
@@ -1,5 +1,6 @@
 from astropy import units as u
 from functools import cached_property
+import numpy as np
 
 from ipyvuetify import VuetifyTemplate
 from glue.core import HubListener
@@ -222,7 +223,14 @@ class FluxColumnSelect(SelectPluginComponent):
         # manipulate the arrays in the data-collection directly, and modify FLUX_ORIGIN so that
         # exporting back to a lightkurve object works as expected
         self.app._jdaviz_helper._set_data_component(dc_item, 'flux', dc_item[self.selected])
-        self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err', dc_item[self.selected+"_err"])  # noqa
+        if self.selected+"_err" in [c.label for c in dc_item.components]:
+            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err',
+                                                        dc_item[self.selected+"_err"])  # noqa
+        else:
+            nan_errs = np.empty(dc_item['flux'].shape)
+            nan_errs[:] = np.nan
+            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err', nan_errs)
+
         dc_item.meta['FLUX_ORIGIN'] = self.selected
 
         self.hub.broadcast(FluxColumnChangedMessage(dataset=self.dataset.selected,
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -175,6 +175,7 @@ class LCviz(ConfigHelper):
                                   parser_reference='tess_dvt_parser',
                                   data_label=data_label,
                                   extname=extname)
+                return
 
         super().load_data(
             data=data,
--- a/lcviz/helper.py
+++ b/lcviz/helper.py
@@ -1,3 +1,4 @@
+from astropy.io.fits import getheader
 import astropy.units as u
 import ipyvue
 import os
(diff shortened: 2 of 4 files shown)


rosteen:
I think this is ready for testing/opinions on how it's implemented, I imagine I'll need to iterate a little.

lcviz/parsers.py
```diff
@@ -12,9 +14,46 @@
...
+    extname = kwargs.pop('extname')
+
+    # Loop through the TCEs in the file. If we only want one (specified by
+    # `extname` keyword) then only load that one into the viewers and ephemeris.
+    for i in range(1, len(hdulist)-1):
+        data = hdulist[i].data
+        header = hdulist[i].header
+        lc = lightkurve.LightCurve(data=data,
+                                   flux=data['LC_INIT'],
+                                   flux_err=data['LC_INIT_ERR'])
```
bmorris3, line 37: We have some work to do on the units here. We're currently expecting the time unit to be in JD (I think?), but these files have the units:
```
TTYPE1  = 'TIME    '           / column title: data time stamps                 
TFORM1  = 'D       '           / column format: 64-bit floating point   
TUNIT1  = 'BJD - 2457000, days' / column units: Barycenter corrected TESS Julian
```
but the offset of `2457000` isn't being reapplied in this parser yet, so the time axis is prehistoric:

Screenshot 2025-01-02 at 10 05 51


We might also need to intentionally _not_ load the DVT's phase column, or mark its caveats. The phase units are not the (0, 1) orbital phase that we usually use:
```
TTYPE4  = 'PHASE   '           / column title: Phase using period and epoch     
TFORM4  = 'E       '           / column format: 32-bit floating point           
TUNIT4  = 'days    '           / column units: [-0.25*period, 0.75*period]     
```

lcviz/parsers.py
```diff
@@ -12,9 +14,46 @@
...
+        lc.meta = hdulist[0].header
+        lc.meta['MISSION'] = 'TESS DVT'
+        lc.meta['FLUX_ORIGIN'] = "LC_INIT"
```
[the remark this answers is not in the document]
bmorris3: 🤔  Down the road, it'll be great to change the plot options defaults for "MODEL_*" columns.

lcviz/parsers.py
kecnry: agreed, we probably shouldn't load the phase column but could load the ephemeris and let lcviz automatically phase-fold (if that's easy - otherwise can be a follow-up ticket)

rosteen: > agreed, we probably shouldn't load the phase column but could load the ephemeris and let lcviz automatically phase-fold (if that's easy - otherwise can be a follow-up ticket)

I already made a pass at doing that at L53-54.

lcviz/parsers.py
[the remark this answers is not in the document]
rosteen: I think I fixed this in `components.py` but it needs more testing to make sure something downstream isn't broken by an all-nan uncertainty array.

lcviz/helper.py
```diff
@@ -161,7 +162,21 @@ def load_data(self, data, data_label=None):
             automatically determined from filename or randomly generated.
             The final label shown in LCviz may have additional information
             appended for clarity.
+        extname : str or `None`
+            Used for DVT parsing if only a single TCE from a multi-TCE file should be
+            loaded. Formatted as 'TCE_1', 'TCE_2', etc.
         """
+        # Determine if we're loading a DVT file, which has a separate parser
+        if isinstance(data, str):
```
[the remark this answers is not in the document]
rosteen: I agree, but handling HDULists in general is probably a separate ticket/PR.

rosteen:
@bmorris3  I fixed the time handling, but I'm surprised that using the epoch and period from the DVT header doesn't seem to result in a nice light curve in the time vs phase viewer:

Screenshot 2025-01-02 at 1 50 58 PM

rosteen:
Ah ha, I switched to the model and realized that the period is good, it's just not centered at phase 0.5. Here's the folded transit:

Screenshot 2025-01-02 at 2 31 03 PM

lcviz/parsers.py
```diff
@@ -12,9 +16,54 @@
...
+        else:
+            show_ext_in_viewer = show_in_viewer
+
+        light_curve_parser(app, lc, data_label=data_label,
+                           show_in_viewer=show_ext_in_viewer, **kwargs)
+
+        # add ephemeris information from the DVT extension
+        if ephem_plugin is not None and show_ext_in_viewer:
+            ephem_plugin.period = header['TPERIOD']
+            ephem_plugin.t0 = header['TEPOCH'] + time_offset
```
bmorris3, line 64: The midtransit time doesn't align with this epoch because of the time conventions – when we get it right, the midtransit time should be phase zero.

You're correctly applying the BJD offset to `TEPOCH`, but then you need to subtract the LC's coordinate reference time like this:
```python
dvt_file = "https://archive.stsci.edu/missions/tess/tid/s0002/0000/0001/0010/0827/tess2018235142541-s0002-s0002-0000000100100827-00109_dvt.fits"

with fits.open(dvt_file, mode='readonly') as hdulist:
    print(
        "t0 for ephemeris plugin:", 
        (hdulist[1].header['TEPOCH'] + 2457000) - 
        lcviz.app.data_collection[0].coords.reference_time.jd
    )
```
That puts the transit at phase zero.

lcviz/parsers.py
```diff
@@ -12,9 +16,54 @@
...
+    '''
+    hdulist = fits.open(file_obj)
+    ephem_plugin = app._jdaviz_helper.plugins.get('Ephemeris', None)
+    extname = kwargs.pop('extname')
+
+    # Loop through the TCEs in the file. If we only want one (specified by
+    # `extname` keyword) then only load that one into the viewers and ephemeris.
+    for i in range(1, len(hdulist)-1):
+        data = Table(hdulist[i].data)
+        # We don't want these columns
```
bmorris3, line 36:
```suggestion
        # don't load some columns with names that may 
        # conflict with components generated later by lcviz
```

lcviz/components/components.py
```diff
@@ -222,7 +223,14 @@ def _on_change_selected(self, *args):
...
         # exporting back to a lightkurve object works as expected
         self.app._jdaviz_helper._set_data_component(dc_item, 'flux', dc_item[self.selected])
-        self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err', dc_item[self.selected+"_err"])  # noqa
+        if self.selected+"_err" in dc_item.component_ids():
+            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err',
+                                                        dc_item[self.selected+"_err"])  # noqa
+        else:
+            nan_errs = np.empty(dc_item['flux'].shape)
+            nan_errs[:] = np.nan
+            self.app._jdaviz_helper._set_data_component(dc_item, 'flux_err', nan_errs)
```
bmorris3, line 232: Can we explicitly add support for light curves without flux uncertainties, rather than adding nan uncertainties? If/when we add fitting features, we'll have to undo the nans.

kecnry: What currently requires uncertainties?  I can't think of anywhere that we currently use them (although that really should be bumped up near the top of our priority list) 🤔

rosteen: So, we can't change the `flux_err` component to be `None`, because you can't change the array shape of a component. We could remove the `flux_err` component entirely in this case and add it back when needed, does that sound better than setting the array to NaNs?

rosteen: Right, there's `LC_INIT` with `LC_INIT_ERR`, but for example `LC_WHITE` doesn't have an `LC_WHITE_ERR` column. We'll need a check either way when fitting is added, either for whether the error exists or if it's all NaN, just a matter of which to check. Sounds like @kecnry and I don't have a strong preference so I can defer to @bmorris3 and remove the component instead of doing all NaN.

bmorris3 (approved):

lcviz/parsers.py
```diff
@@ -12,9 +16,55 @@
...
+    '''
+    Read a TESS DVT file and create a lightkurve object
+    '''
+    hdulist = fits.open(file_obj)
+    ephem_plugin = app._jdaviz_helper.plugins.get('Ephemeris', None)
+    extname = kwargs.pop('extname')
+
+    # Loop through the TCEs in the file. If we only want one (specified by
+    # `extname` keyword) then only load that one into the viewers and ephemeris.
+    for i in range(1, len(hdulist)-1):
```
kecnry, line 34: can we just iterate over hdulist directly (`for hdulitem in hdulist[1:]`) or is the iterator `i` needed anywhere else?

kecnry (commented):
Just one small comment, otherwise looks good to me.  Are there any necessary follow-ups or does this at least achieve everything requested by MAST?

lcviz/parsers.py
rosteen: I think at some point I was going to accept an integer rather than string name for extension, I can change this to what you suggested.

rosteen:
> Just one small comment, otherwise looks good to me. Are there any necessary follow-ups or does this at least achieve everything requested by MAST?

I think that this covers everything MAST needs, if they decide they want to provide an integer instead of `extname` for TCE event it would be a small change. I think we can get it into their hands like this for testing on their end.

PR #174, opened 2025-02-28
fix link to lightkurve docs

ttdu:
docs.lightkurve.org is the old link, which is now broken -- this updates to the latest link


kecnry:
Thanks @ttdu!

See also #144 (both of these should be reverted if they fix the original docs location) - see https://github.com/lightkurve/lightkurve/issues/1466

kecnry (approved):
LGTM - failing tests are unrelated

PR #180, opened 2025-03-25
Fix stitch data removal with Jdaviz 4.2

rosteen:
Jdaviz no longer has an `app.vue_data_item_remove` method, so this needed to be updated to use the backend method.


kecnry (approved):
Thanks!

kecnry:
Should we just add this to the existing update to 4.2 changelog entry or skip changelog?

PR #182, opened 2025-04-03
parser: do not require TELESCOP to be set

kecnry:
This PR allows files to be parsed where `TELESCOP` is not set.


kecnry:
@meeseeksdev backport to v1.1.x

PR #197, opened 2025-07-10
Updated documentation reference in preparation for PR #3685 in jdaviz

MatthewPortman:
* https://github.com/spacetelescope/jdaviz/pull/3685

**Waiting for**: jdaviz 4.3 released and pinned via #183


kecnry:
ah, forgot this existed, but ended up having to do this before updating the pin (see #203)

PR #215, opened 2026-04-10
jdaviz 5.0 updates

kecnry:
This PR update jdaviz to be compatible with the deconfigged jdaviz (5.0 release).

This requires https://github.com/spacetelescope/jdaviz/pull/4135 (currently in jdaviz main, will be in jdaviz 5.0.2 once release - this cannot be merged until jdaviz is released and pinned here).

This deprecates the lcviz config itself, and instead suggests injecting capability into jdaviz ~(docs updates still to come - see #217)~:

```
import jdaviz as jd

lc = search_lightcurve("HAT-P-11", mission="Kepler",
                       cadence="long", quarter=10).download()
jd.show()
```

```
ldr = jd.loaders['object']
ldr.object = lc
ldr.open_in_tray()
# will only show "Catalog" as a format option
```

```
import lcviz
# will now show light curve loaders/importers
ldr.format = "Light Curve"
```

```
ldr.load()
```


pyproject.toml
```diff
@@ -20,9 +20,8 @@ classifiers = [
 ]
 dependencies = [
     "astropy>=5.2",
-    # NOTE: if/when we stop pinning a minor version of jdaviz, add jdaviz
-    # to devdeps in tox.ini
-    "jdaviz>=4.5,<4.6",
+    # TODO: update to jdaviz>=5.0 once jdaviz 5.0 is released
+    "jdaviz @ git+https://github.com/spacetelescope/jdaviz.git@main",
```
kecnry: NOTE: need to update to `>5.0.2,<5.1` after jdaviz 5.0.2 is released and before this is merged

lcviz/tools.py
```diff
@@ -9,4 +10,16 @@
 SidebarShortcutExportPlot.plugin_name = 'lcviz-export'
 
 
+# Override SelectSlice tool visibility to include 1D light curves
+def _selectslice_is_visible(self):
+    # visible if any 1D light curve or 3D cube-like data is present
+    for dc in self.viewer.jdaviz_app.data_collection:
+        if dc.ndim in (1, 3):
+            return True
```
rosteen, line 18: This may need to be a bit more specific in order to differentiate between spectra and light curves.

rosteen:
I was able to replicate the old example workflow with this + Jdaviz 5.x, so in that sense it's a success. But I started digging into loading different types of data along with light curves and immediately noticed that the slice tool interacts with spectra+light curves in ways we don't want:

https://github.com/user-attachments/assets/143657d3-4355-4ff0-8a68-4e57311289eb

Spectral subsets also show up in the time series viewers as layers (although nothing appears to be highlighted). Maybe mixed data workflows can be a follow up since this already preserves the existing capabilities.

bmorris3:
## Comments about this PR

* this is great. it worked for me when I experimented with the usual lcviz examples. I didn't load other data types to check for issues like Ricky's.

## Comments that have nothing to do with this PR
(don't address these in this PR. just notes for later.)


* can we remove the "show cube viewer" button from time selector in this context? It opens an empty image viewer, but no TPF is loaded, so no cube(/slice) is shown. 
Screenshot 2026-06-09 at 16 00 07


* I've run the frequency analysis plugin's BLS method in units of period (popout window on the right in the screenshot below). The ephemeris plugin identifies the period at maximum power at 0.36 d. Could we/should we allow the user to set the range in the ephemeris plugin, like we do in frequency analysis?

Screenshot 2026-06-09 at 16 09 02

* we should create ra/dec `Component`s on the light curve `Data` entries, so we could link a source catalog displayed on an image with the light curves.

kecnry:
@rosteen - I think I fixed that case in the latest commit here and in https://github.com/spacetelescope/jdaviz/pull/4225 

@bmorris3 - I'll create tickets for the first 2 points you made (I think the last one is already effectively planned for upcoming MOS-workflows, but definitely something we will need to add here in lcviz as well once we settle exactly how its going to look in jdaviz first).

bmorris3 (approved):
Looks good. Absurd that the `oktopus` issue hasn't been addressed yet.

PR #217, opened 2026-04-21
TST: wireframe demo

kecnry:
test inheriting the wireframe demo from jdaviz

this PR should not be merged, but rather merged into #215


kecnry:
this is now a part of #215

PR #220, opened 2026-06-08
lightkurve resolver

kecnry:
~This should be rebased on main after #215 is merged~

This implements a new resolver (loader source) injected into jdaviz when lcviz is imported to query lightkurve.


lcviz/loaders/resolvers/lightkurve/lightkurve.vue
```diff
@@ -0,0 +1,180 @@
...
+        ></v-switch>
+      </v-row>
+
+      <v-row>
+        <div :style="!(viewer_selected !== 'Manual' && !coord_follow_viewer_pan) ? 'width: 100%' : 'width: calc(100% - 32px)'">
+          <v-text-field
+            v-model="source"
+            :label="api_hints_enabled ? 'ldr.source =' : 'Source/Coordinates'"
+            :class="api_hints_enabled ? 'api-hint' : null"
+            hint="Enter a source name or coordinates in degrees to center your query on"
```
bmorris3:
```suggestion
            hint="Enter a source name or source coordinates in degrees"
```

lcviz/loaders/resolvers/lightkurve/lightkurve.vue
```diff
@@ -0,0 +1,180 @@
...
+        </div>
+      </v-row>
+
+      <plugin-select
+        :items="coordframe_choices.map(i => i.label)"
+        :selected.sync="coordframe_selected"
+        label="Coordinate Frame"
+        api_hint="ldr.coordframe ="
+        :api_hints_enabled="api_hints_enabled"
+        hint="Astronomical Coordinate Frame of the provided Coordinates"
```
bmorris3:
```suggestion
        hint="Astronomical coordinate frame for the query coordinate"
```

bmorris3 (approved):
I'm approving this PR, and I'm leaving non-blocking comments below.

### Comments

* I often use the `cadence`/`exptime` and `quarter/sector` kwargs. Let's make a follow-up ticket for those. Even if we don't support quarter/sector, I think there should be a toggle/button to "query and load all" or "query, stitch all, then load". I recognize the latter is asking for a simple solution to an underdefined workflow.

* The spinner on "query archive" is great. I think we need a progress bar/spinner for the "Importer" section. It took maybe 1.5 - 3 seconds after I selected a file for the Importer section to be populated. The (temporarily) empty section had me concerned something broke. This comment shouldn't block merge.

* For a future jdaviz PR – the lightkurve resolver appears last in the list, but the vertical height of the dropdown does not stretch to include the last element, making it difficult to discover:

https://github.com/user-attachments/assets/df039368-0573-464d-90d2-abf36178b578

kecnry:
> The spinner on "query archive" is great. I think we need a progress bar/spinner for the "Importer" section. It took maybe 1.5 - 3 seconds after I selected a file for the Importer section to be populated. The (temporarily) empty section had me concerned something broke. This comment shouldn't block merge.

I think this should eventually come for free from https://github.com/spacetelescope/jdaviz/pull/4203 (once we update to jdaviz 5.1)

Created tickets for your other 2 ideas - thanks!

